PR #201, opened 2015-10-12
Reset main.jsbundle to match the previously built SDK

lental:
main.jsbundle has to match the ooyalaSkinSDK that is hard-built into the sample application. It cannot be updated until we update the ooyalaSkinSDK

We are not able to update this file between builds.  This may be a good reason to move it back into the sample applicaiton


lental:
updated sample app to directly reference the vendor/Ooyala/OoyalaSkinSDK/main.jsbundle.  This way we can keep modifying the sdk/OoyalaSkinSDK/iOS/main.jsbundle without destroying the sample app in the repo

PR #206, opened 2015-10-14
[Pba 2553] Loading spinner not shown after IMA mid rolls

rcvrgs:
Do we really need to have the rate in the if condition for the _renderLoading function (of VideoViewPlayPause.js)?

If I remove it, the problem is fixed, as I update the 'rate' everytime we get an onTimeChanged notification from the player, but the 'rate' is always 0 for the player when the ads are playing. I don't know if this is correct, but if I leave rate in the if condition in _renderLoading, then the loading spinner shows when ads are playing, which get us a new bug.

So, what do you think?

### DIFF
--- a/sdk/OoyalaSkinSDK/widgets/VideoViewPlayPause.js
+++ b/sdk/OoyalaSkinSDK/widgets/VideoViewPlayPause.js
@@ -132,7 +132,7 @@ var VideoViewPlayPause = React.createClass({
   },
 
   _renderLoading: function() {
-    if((this.props.rate <= 0 || this.props.playhead == 0) && !this.props.showButton && !this.state.controlPlaying) {
+    if(this.props.playhead == 0 && !this.props.showButton && !this.state.controlPlaying) {
       return (
         <View style={styles.loading}>
           <ActivityIndicatorIOS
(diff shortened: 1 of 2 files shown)


sdk/OoyalaSkinSDK/widgets/VideoViewPlayPause.js
```diff
@@ -132,7 +132,7 @@ var VideoViewPlayPause = React.createClass({
   },
 
   _renderLoading: function() {
-    if((this.props.rate <= 0 || this.props.playhead == 0) && !this.props.showButton && !this.state.controlPlaying) {
+    if(this.props.playhead == 0 && !this.props.showButton && !this.state.controlPlaying) {
```
lental: this won't work, because it effectively won't show the loading spinner at all except for the beginning (when playhead is 0)

we'll need to think of a better way to ensure the loading spinner is only showing when we want it to

rcvrgs: Ok, then I'll explore alternative ways to do it. Now I know rate is necessary in the renderLoading func.

rcvrgs:
@lental, should I close this PR? We'll need to recreate it when we finally fix it.

lental:
yeah, close it off

PR #208, opened 2015-10-16
[PBA-2530] Playback works in iOS7

rcvrgs:
For some reason I don't know yet, the JS for (... of ...) syntax isn't recognizable in iOS7. I changed it to a basic for loop.


lental:
good catch. for... of... is pretty new stuff: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of

the JavascrptCore of iOS 7 probably doesn't have that available

PR #209, opened 2015-10-16
[Pba 2657] Share Panel with localizable string

rcvrgs:
A thing to note, I upgraded the skin-config submodule to a more recent branch. Didn't upgrade to the most recent. What do you think? Should we upgrade to the latest?


lental:
will read in a sec

lental:
any forward progress in the skin-config is good.  If you don't want to do the full update to the skin-config, don't worry about it.  As long as the commit you've chosen is a commit on 'master' branch.

lental:
LGTM

PR #212, opened 2015-10-21
[Logging] Add Log class which can be used to help toggle logging levels

lental:
It is generally first set in index.ios.js, but I can be convinced to move it elsewhere


lental:
bump plz

PR #213, opened 2015-10-22
Pba 2466

soapyigu:


sdk/OoyalaSkinSDK/index.ios.js
```diff
@@ -121,6 +121,9 @@ var OoyalaSkin = React.createClass({
         break;
       case BUTTON_NAMES.PLAY_PAUSE:
         this.setState({showPlayButton: !this.state.showPlayButton});
+      case BUTTON_NAMES.LEARNMORE:
```
lental: This information should come back due to the "pause" in the handle learn more

PR #224, opened 2015-10-28
[PBA-2671] Adapt control bar's duration string

brol1dev:
If the control bar's width is <= 200, the duration string will be smaller, so all the controls are visible.


sdk/OoyalaSkinSDK/controlBar.js
```diff
@@ -60,7 +60,11 @@ var ControlBar = React.createClass({
     if (this.props.live) {
       return this.props.live.label;
     } else {
-      return Utils.secondsToString(this.props.playhead) + "/" + Utils.secondsToString(this.props.duration);
+      var durationString = Utils.secondsToString(this.props.playhead);
```
lental: there are better ways to do this.  Any chance we can use the responsiveDesignManager to help?

lental: hmm maybe not.  as long as you can tell me this is done in the same way "adBar" is done

brol1dev: I don't know how the adBar does it. Will take a look.

sdk/OoyalaSkinSDK/collapsingBarUtils.js
```diff
@@ -29,7 +29,10 @@ var CollapsingBarUtils = {
 
   _collapse: function( barWidth, orderedItems ) {
     var r = { fit : orderedItems.slice(), overflow : [] };
-    var usedWidth = orderedItems.reduce( function(p,c,i,a) { return p+c.minWidth; }, 0 );
+    var usedWidth = orderedItems.reduce( function(p,c,i,a) {
+      if (c.minWidth) return p+c.minWidth;
```
brol1dev, line 33: This is the important line that fixed the problem.

brol1dev:
I moved some utility functions in OOSkinViewController.m to their own utility files. At first, I did it thinking I was going to make a module that exports those utility functions to JS, but at the end the fix was in the collapse functions in JS.

Another thing, I think the duration string can be shown in the control bar, as a simpler version (as I introduced in my first commit), but this would change the code base a bit. Atm, the bug is out, that's the important thing.

lental:
good call, i understand the bug that is fixed, but I would be curious why you had to change the collapsing util that way.  Any way you can add unit tests that would have caught that bug?

brol1dev:
I'm going to be adding unit test for that js file. I think there's already, but it seems it exist on the same file. The unit test should be in a different location, so everything is modular and maintainable.

Also, I'll add unit test for the util classes created in Obj-C :)

PR #226, opened 2015-10-28
[PBA-2741] fix column button clickable error and remove redundant code

soapyigu:


sdk/OoyalaSkinSDK/MoreOptionScreen.js
```diff
@@ -54,7 +54,7 @@ var MoreOptionScreen = React.createClass({
       Animated.timing(                      
         this.state.translateY,                 
         {
-          toValue: 0,                         
+          toValue: this.props.height / 2 - this.props.config.moreOptions.iconSize,
```
lental: are you sure you dont' mean "(props.height - iconsize)/2?  you're technically not going to the center of the screen with your code

PR #227, opened 2015-10-29
[android-infra] refactor index.os so that core component can be share…

Matthew-C:
…d by both platforms.


sdk/OoyalaSkinSDK/ooyalaSkinCore.js
```diff
@@ -0,0 +1,459 @@
+/**
+ * Sample React Native App
+ * https://github.com/facebook/react-native
+ */
+'use strict';
+
+var React = require('react-native');
+var {
+  ActivityIndicatorIOS,
```
lental, line 9: is it okay for these to be part of the core? specifically, iOS specific elements

PR #231, opened 2015-10-30
[PBA-2488] toggle play/pause when IMA ad is tapped.

Matthew-C:


sdk/OoyalaSkinSDK/OoyalaSkinSDK/OOSkinViewController.m
```diff
@@ -274,6 +275,16 @@ - (void) bridgeAdStartNotification:(NSNotification *)notification {
   }
 }
 
+- (void) bridgeAdTappedNotification:(NSNotification *)notification {
```
lental, line 278: huuuh.  much cooler than i was originally thinking.  I'll see how it looks either today or monday.  awesome!

lental, line 278: Can you add a comment in OOSkinViewController for the bridgeAdTappedNotification, to communicate to others when that should be expected (Only IMA)

lental:
this was pretty sweet. thix change is strictly better than what we currently have.  However, here are two other things which should be fixed later:

The control bar should probably not show during the IMA Ad, if possible
The PlayPause buttons till can be put into the "Buffering Logo" state and unable to leave.  

Both of these are not expected for this ticket, so I can merge after the addition of the comment

Matthew-C:
comment is added. I'll hide the control bar and try to fix loading logic (in a separate checkin).

lental:
:+1:

PR #232, opened 2015-10-30
[PBA-2671] avoid showing the red-screen error when ignoring hidden op…

lental:
…tions in the More Options


brol1dev:
Oh, so when you do a console.error the red screen appears. Good to know.

PR #234, opened 2015-11-02
[Pba 2753] Control bar not visible on end screen when discovery is on

brol1dev:
Control bar still shows when discovery is off.

### DIFF
--- a/sdk/OoyalaSkinSDK/EndScreen.js
+++ b/sdk/OoyalaSkinSDK/EndScreen.js
@@ -65,8 +65,6 @@ var EndScreen = React.createClass({
     return (
       <View style={styles.fullscreenContainer}>
         {this.props.discoveryPanel}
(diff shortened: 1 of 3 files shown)


sdk/OoyalaSkinSDK/EndScreen.js
```diff
@@ -65,8 +65,6 @@ var EndScreen = React.createClass({
     return (
       <View style={styles.fullscreenContainer}>
         {this.props.discoveryPanel}
-        {progressBar}
-        {controlBar}
```
lental, line 69: you're sure we can just remove these?

brol1dev, line 69: Those lines add the control bar, I think it's ok to remove them. But let's chat about what is the expected behavior.

PR #238, opened 2015-11-03
[FixTesting] remove now obsolete test script. Use 'npm test' instead.

jonslenkooyala:


brol1dev:
Ok, I updated jest-cli on this branch ;)

mistermendez:
Apparently the latest v4 versions of node.js work with the latest version of jest (that's using io.js)

PR #242, opened 2015-11-04
[PBA-2744] Share success pop up doesn't appear at start anymore

brol1dev:
The problem was that after sending a successful message over Twitter or Facebook, we show a pop up telling that the post was successful, but when you exit the share screen to playback the video, and then return to the share screen again, the first thing you would see is the same pop up that message was successfully sent, but we haven't sent anything at this moment.

### DIFF
--- a/sdk/OoyalaSkinSDK/ooyalaSkinCore.js
+++ b/sdk/OoyalaSkinSDK/ooyalaSkinCore.js
@@ -85,6 +85,11 @@ OoyalaSkinCore.prototype.onSocialButtonPress = function(socialType) {
   (results) => {Log.log(results);});
 };
 
+OoyalaSkinCore.prototype.onSocialAlertDismiss = function() {
+  this.skin.setState({alertTitle: ''});
+  this.skin.setState({alertMessage: ''});
+};
+
 OoyalaSkinCore.prototype.pauseOnOptions = function() {
   if (this.skin.state.screenType != SCREEN_TYPES.MOREOPTION_SCREEN) {
     this.previousScreenType = this.skin.state.screenType;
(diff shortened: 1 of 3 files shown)


lental:
good stuff.  will read and test before merging

sdk/OoyalaSkinSDK/ooyalaSkinCore.js
```diff
@@ -85,6 +85,11 @@ OoyalaSkinCore.prototype.onSocialButtonPress = function(socialType) {
   (results) => {Log.log(results);});
 };
 
+OoyalaSkinCore.prototype.onSocialAlertDismiss = function() {
+  this.skin.setState({alertTitle: ''});
+  this.skin.setState({alertMessage: ''});
```
lental, line 90: not a big fan of these state names, but i'll let it pass for now because you didn't add it

sdk/OoyalaSkinSDK/sharePanel.js
```diff
@@ -52,19 +53,24 @@ var SharePanel = React.createClass ({
...
+		if(this.props.alertTitle !== ""){
 			postShareSuccessAlert = AlertIOS.alert(
 				this.props.alertTitle,
-  			this.props.alertMessage,
-  			[
-    			{text: 'Ok', onPress: () => Log.log('Ok Pressed!')},
-  			]
+				this.props.alertMessage,
+				[
+					{ text: 'Ok', onPress: this.onSocialAlertDismiss },
```
lental, line 72: is there any way to click out of the box (in other words, not click on the "Ok" button)?

lental:
this is still strictly better, so i will add it.  however I would like a response for my comment

sdk/OoyalaSkinSDK/sharePanel.js
brol1dev, line 72: I'll check if this is possible and let you know.

PR #245, opened 2015-11-06
[PBA-2802] Add JS listener to volume change event

brol1dev:
I don't know why, but it when the ooyalaSkinCore.js was created, the volume listener I added was removed. I added it again and the volume icon is being shown correctly once more.

Also changed the event name to 'volumeChanged', as it resembles to the others.


lental:
must've been a bad merge.  good catch, will be in the next build

PR #246, opened 2015-11-06
[PBA 2657] Sync localization with skin-config

brol1dev:
I refactored the code a bit, so that OOSkinViewController doesn't have a lot of methods.

The main problem was that the code wasn't updated to the newer changes in skin-config, so localization wasn't working.

I wanted to add unit test for 'utils.js', but I wasn't able to setup that test case with Jest. I'm getting errors, it seems that it isn't mocking some react stuff correctly, as we require React in utils.js.

### DIFF
--- a/sdk/OoyalaSkinSDK/OoyalaSkinSDK/Utils/NSDictionary+Utils.m
+++ b/sdk/OoyalaSkinSDK/OoyalaSkinSDK/Utils/NSDictionary+Utils.m
@@ -8,12 +8,11 @@
 
 #import "NSDictionary+Utils.h"
 
+#import "OOSkinViewController.h"
 #import <OoyalaSDK/OODebugMode.h>
 #import "OOLocaleHelper.h"
 #import "NSMutableDictionary+Utils.h"
 
-static NSString * const kLocalizableStrings = @"localization";
-static NSString * const kLocale = @"locale";
 
 @implementation NSDictionary (Utils)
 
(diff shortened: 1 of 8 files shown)


sdk/OoyalaSkinSDK/OoyalaSkinSDK/Utils/NSDictionary+Utils.m
```diff
@@ -0,0 +1,63 @@
...
+//  Copyright © 2015 ooyala. All rights reserved.
+//
+
+#import "NSDictionary+Utils.h"
+
+#import <OoyalaSDK/OODebugMode.h>
+#import "OOLocaleHelper.h"
+#import "NSMutableDictionary+Utils.h"
+
+static NSString * const kLocalizableStrings = @"localization";
```
lental: can these constants be removed from the OoSkinViewController? Are they there as well?

brol1dev: Oh yea, let me remove them.

brol1dev:
Oh wait @lental, I removed the constants, but didn't check before pushing. The constants are still being used in OOSkinViewController. I guess I'll keep them there and use them from the other class I created.

brol1dev:
It's ready now.

But I think we should find where each of our constants should reside.

lental:
I agree, not suuper happy about having unnecessary things like that in our header, but it's okay for now.

I think our locale helper should just take the skin-config variable, and localehelper itself should have the constants

lental:
actually, do you think you can fix this?  modifying OOLocaleHelper localizedString to  accept the skinconfig, instead of skinconfig[klocalizablestrings]

brol1dev:
Ok, will do.

brol1dev:
Missed that one. OOLocaleHelper has the constants now.

lental:
perfect! thanks

PR #248, opened 2015-11-07
Pba 2553c fix loading logic for playpause widget

Matthew-C:
1. cleanup videoviewplaypause
2. make meaningful method name
3. kick-off animations in the right (in my opinion) place.


lental:
there are still a handful of bugs on this branch.  I want to dig into it, but I may not be able to for now

lental:
example 1: the More Options menu no longer pauses video playback

PR #251, opened 2015-11-13
[PBA-2767] Implement continuos playback on discovery end screen

brol1dev:
First code version for the continuos playback on the discovery end screen. **Don't merge it yet**, I will add some more code to make it more robust, but on this version you can see it in action.

### DIFF
--- a/sdk/OoyalaSkinSDK/discoveryPanel.js
+++ b/sdk/OoyalaSkinSDK/discoveryPanel.js
@@ -47,12 +47,12 @@ var DiscoveryPanel = React.createClass({
   componentDidMount:function () {
     this.state.opacity.setValue(0);
     Animated.parallel([
-      Animated.timing(                      
-        this.state.opacity,                 
+      Animated.timing(
+        this.state.opacity,
         {
-          toValue: 1,                         
+          toValue: 1,
           duration: animationDuration,
-          delay: 0  
+          delay: 0
         }),
     ]).start();
   },
@@ -81,7 +81,7 @@ var DiscoveryPanel = React.createClass({
       thumbnailStyle = styles.thumbnailLandscape;
       columnContainerStyle = styles.columnContainerLandscape;
     }
-    
+
     var animationStyle = {opacity:this.state.opacity};
     return (
       <Animated.View style={[styles.panel, animationStyle]}>
@@ -118,6 +118,7 @@ var DiscoveryPanel = React.createClass({
 
     return (
     <TouchableHighlight
+      key={sectionID}
       underlayColor='#37455B'
       onPress={() => this.onRowSelected(item)}
       style={itemRect}>
--- a/sdk/OoyalaSkinSDK/discoveryPanel.js
+++ b/sdk/OoyalaSkinSDK/discoveryPanel.js
@@ -20,13 +20,16 @@ var Utils = require('./utils');
 var ResponsiveList = require('./widgets/ResponsiveList');
 var CircularStatus = require('./widgets/CircularStatus');
 var styles = Utils.getStyles(require('./style/discoveryPanelStyles.json'));
+var Constants = require('./constants');
+var {
+  SCREEN_TYPES,
+} = Constants;
 // TODO: read this from config.
 var itemRect;
 var thumbnailStyle;
 var columnContainerStyle;
 var widthPortrait = 375;
 var animationDuration = 1000;
-
 
 var DiscoveryPanel = React.createClass({
   propTypes: {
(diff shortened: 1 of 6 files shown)


brol1dev:
You can merge now if you want :)

sdk/OoyalaSkinSDK/discoveryPanel.js
```diff
@@ -69,6 +81,30 @@ var DiscoveryPanel = React.createClass({
...
+      showCircularStatus: true,
+    });
+  },
+
+  updateTimer: function(row) {
+    if (this.state.currentCounterVal == 0) {
+      this.onRowSelected(row);
+    } else {
+      var self = this;
+      setTimeout(function() {
```
lental, line 101: This is only slightly concerning to me - It is the UI causing things to happen.  I don't have the best idea of if there is a better way to do this, but I'll voice my concern at least, and keep reading

sdk/OoyalaSkinSDK/discoveryPanel.js
```diff
@@ -99,6 +135,16 @@ var DiscoveryPanel = React.createClass({
...
   },
 
+  renderCircularStatus: function() {
+    return (
+      <CircularStatus
+        onPress={() => this.onStatusPressed()}
+        total={this.state.counterLimit}
+        current={this.state.currentCounterVal}
+        thickness={2}
+        diameter={44} />);
```
lental, line 145: we have to worry about hard coded values at some point, maybe not at this point

sdk/OoyalaSkinSDK/widgets/CircularStatus.js
```diff
@@ -17,10 +17,11 @@ var CircularStatus = React.createClass({
     current: React.PropTypes.number,
     total: React.PropTypes.number,
     diameter: React.PropTypes.number,
-    thickness: React.PropTypes.number
+    thickness: React.PropTypes.number,
```
lental, line 20: why did you add thickness?

sdk/OoyalaSkinSDK/widgets/ResponsiveList.js
```diff
@@ -67,20 +67,25 @@ var ResponsiveList = React.createClass({
 
   renderSlice: function(slice: object, i: number) {
     var sliceStyle = this.props.horizontal ? styles.column : styles.row;
+
+    renderItem = this.renderItem;
```
lental: is there any reason you need this as a variable?

sdk/OoyalaSkinSDK/discoveryPanel.js
brol1dev, line 101: Ok, let me think of another way of implementing this.

sdk/OoyalaSkinSDK/widgets/CircularStatus.js
brol1dev, line 20: I just added the comma at the end.

lental, line 20: whoops, true

sdk/OoyalaSkinSDK/widgets/ResponsiveList.js
brol1dev: I did that because if we call 'this' inside the slice.map context, we can't call the defined functions from the ResponsiveList. So I make the renderItem a variable so I can access it within slice.map.

sdk/OoyalaSkinSDK/discoveryPanel.js
lental, line 101: i wouldn't say dump it yet.  I think it's okay, and if you have any new interesting ideas, let me know before coding it up.  it's more of a curious case to think about

brol1dev, line 101: I'm thinking of having a module (maybe residing within the CircularStatus.js) that would send an event to the UI (discoveryPanel.js) that a new second thick in the clock happened. So the UI is only responsible of updating the UI by receiving this event. But I really don't know how hard it would be to implement it.

sdk/OoyalaSkinSDK/widgets/ResponsiveList.js
lental: is it possible to add the 'var' to the front of it?  it feels like this is leaking a little out of scope

brol1dev: Oh yea, totally. I missed that one.

sdk/OoyalaSkinSDK/discoveryPanel.js
brol1dev, line 145: I agree, but where could be a good place for they to live? Within constants? In the skin-config?

sdk/OoyalaSkinSDK/discoveryPanel.js
lental, line 101: First, do you need to stop the timeout at any point, or does this go on forever?

I think its fine how it is currently.  I think i may be overcomplicating UI. @Matthew-C , do you have any opinions on this? using setTimeout in javascript for timers, and then using it to autoplay the next video?

lental:
Gonna let it in for now. We can discuss changes going forward. thanks

PR #254, opened 2015-11-18
[PBA-2767] Fix timeout usage

brol1dev:
setTimeout was being called more than once. We may need to eliminate setTimeout completely, but for now this works. I added React's recommended Timer mixin, that substitute the global setTimeout with their own, so is safer to use on React native.

One thing I don't like with this timeout implementation, is that we get a warning in the console each time the timer is updated: 'Warning: setState(...): Cannot update during an existing state transition (such as within `render`). Render methods should be a pure function of props and state.' We may need to think in another way to update the timer, as they way I'm doing it, I'm updating a var that triggers a re-render of the UI within the render function, and that shouldn't happen.

### DIFF
--- a/sdk/OoyalaSkinSDK/discoveryPanel.js
+++ b/sdk/OoyalaSkinSDK/discoveryPanel.js
@@ -52,7 +52,7 @@ var DiscoveryPanel = React.createClass({
       showCircularStatus: false,
       currentCounterVal: 0,
       counterLimit: 0,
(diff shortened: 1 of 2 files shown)


brol1dev:
Ok, way better implementation now. Instead of calling the set state within the render function, I start calling an interval when the component mounts. This way, the render functions is only reading state, and not changing it.

PR #258, opened 2015-11-30
Pba 2711

soapyigu:


lental:
nope, this is not what you want to do.  don't merge into stable.  merge into dev.

PR #259, opened 2015-11-30
[PBA-2711] hide the bottomOverlay when the ad is playing

soapyigu:


sdk/react/videoView.js
```diff
@@ -71,6 +71,16 @@ var VideoView = React.createClass({
     locale: React.PropTypes.string
   },
 
+  componentWillReceiveProps: function(nextProps) {
```
lental, line 74: @rcvrgs  @Matthew-C @wunnath @sveravelli  just a heads up on using this -  We should never be modifying state in renders or most other methods.   If state relies on incoming props, use componentWillReceiveProps

PR #261, opened 2015-12-01
[PBA-2859] Smoother countdown timer

brol1dev:
The countdown timer for up next, and discovery screens is reimplemented, it has a smoother animation now :)


brol1dev:
My bad, created PR for stable branch. Will recreate it for dev.

PR #262, opened 2015-12-01
[PBA-2859] Smoother countdown timer

brol1dev:
Re-implemented countdown timer that is being used in the up next and discovery screens.


brol1dev:
At the moment this one only works for iOS. As I use Obj-C code. I'll be working on the Android alternative as time permits :)

PR #253, opened 2015-11-18
changes of startScreen, VideoView , PlayPauseVideoViewAndroid, PauseS…

wunnath:
…creen, EndScreen, index.ios(adding platform support)

PR #269, opened 2015-12-02
[PBA-2878] Rollback scrubber handleTouchMove()

brol1dev:
The PR #253 introduced an additional line of code in the handleTouchMove function
for the bottomOverlay component. I'm removing it as it is causing the bug described
in PBA-2878.

We may put it back again but considering that it breaks the UX in iOS when you try
to drag the scrub bar for the volume component, as it also drags the progress bar.


lental:
@wunnath  and @sveravelli for context.  this may break Android code, but iOS has a regression now, so this needs to be reverted

PR #271, opened 2015-12-03
[PBA-2853] UI fixes to Up next component

brol1dev:
Things to notice:
- I added the Roboto font to the project. If this PR is accepted we'll need to create another one in the ios-sample-apps repo, to install the font in there. I tried it locally and it works.
- I'm not fully respecting the measures that are outlined in the mocks, because if I do, the UI wouldn't look good, so I'm trying my best to use the pts/pixels that I think are the correct ones.


brol1dev:
I already added multiple lines wrap for the title. With this, you can merge the PR now.

I didn't add the ellipsis one, because that one is going to be harder, as we need to know where exactly we need to cut the text. We can know how long (width) the text view is, but I haven't found out where do we cut the text to add the ellipsis, we may thing about this in another iteration if we still need the ellipsis.

At the moment check the multiple line solution.

lental:
Merge conflicts -  sorry, its a hazard of all these open PRs

brol1dev:
Ok, i'll fix it :)

brol1dev:
@lental, I merged dev into this branch. Can merge automatically now.

PR #274, opened 2015-12-04
[PBA-2856] Add close button to Discovery End Screen, move renderButto…

soapyigu:
…n function to Utils.js


sdk/react/utils.js
```diff
@@ -9,10 +9,24 @@ var {
   StyleSheet
 } = React;
 
-
 var Log = require('./log');
 var Utils = {
 
+  renderButton: function(style, icon, func, size, color, fontFamily) {
```
lental: good idea, @Matthew-C any objections to this?

If not, I would rename it to be more specific ( "renderRectButton")

Matthew-C: sure, rename it to renderRectButton

PR #280, opened 2015-12-04
[PBA-2855] resize discovery panel according to device size

soapyigu:


sdk/react/discoveryPanel.js
```diff
@@ -95,14 +94,14 @@ var DiscoveryPanel = React.createClass({
...
     var panelHeight = this.props.height - 40;
 
-    if(this.props.width <= widthPortrait){
-      itemRect = {width: 186, height:164};
-      thumbnailStyle = styles.thumbnailPortrait;
-      columnContainerStyle = styles.columnContainerPortrait;
-    }else{
-      itemRect = {width: 166, height:154};
+    if (this.props.width > this.props.height) {
+      itemRect = {width: this.props.width / 4, height:154};
```
lental: sorry, you should consider doing something with responsiveviewmanager

brol1dev:
I tested it on the iPhone 6 Plus simulator, and I feel it is still a bit more to the left and not centered. Could you confirm if you are looking at this same behavior?

brol1dev:
nice, it looks great now. All centered. Thanks Yi.

PR #282, opened 2015-12-08
fixing play pause and using the same old file of IOS

wunnath:


Matthew-C:
cool stuff. we consolidate playpausevideoview.

PR #283, opened 2015-12-09
[PBA-2409] Adapt discovery scroll direction depending on device orientation

brol1dev:
Discovery scroll direction will change depending on the device orientation. Added a bit of unit testing to utils.js.

### DIFF
--- a/sdk/react/discoveryPanel.js
+++ b/sdk/react/discoveryPanel.js
@@ -93,10 +93,7 @@ var DiscoveryPanel = React.createClass({
   },
 
   render: function() {
-    var renderHorizontal = false;
-    if (Utils.shouldShowLandscape(this.props.width, this.props.height)) {
-      renderHorizontal = true;
-    }
+    var renderHorizontal = Utils.shouldShowLandscape(this.props.width, this.props.height);
 
     var panelHeight = this.props.height - 40;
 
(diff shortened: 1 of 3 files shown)


sdk/react/utils.js
```diff
@@ -28,6 +28,12 @@ var Utils = {
   },
 
   shouldShowLandscape: function(width, height) {
+    if (isNaN(width) || isNaN(height) ||
+        width === null || height === null ||
+        width < 0 || height < 0) {
+      return null;
```
Matthew-C: should this return a boolean?

sdk/react/discoveryPanel.js
```diff
@@ -93,6 +93,11 @@ var DiscoveryPanel = React.createClass({
   },
 
   render: function() {
+    var renderHorizontal = false;
+    if (Utils.shouldShowLandscape(this.props.width, this.props.height)) {
+      renderHorizontal = true;
```
Matthew-C: can we just use var renderHorizontal = Utils.shouldShowLandscape and returns true/false from shouldShowLandscape?

brol1dev: You're right

sdk/react/utils.js
brol1dev: I added null, because if one of those conditions is met, it means that there's something wrong with the parameters. Returning false here, is telling that it shouldShowLandscape == false, but logically that's not true, because by returning null, I'm telling parameters are bad. 

But if we don't want to think about this problem, we may just return false, but if someone passes a bad parameter to this function, it may be troublesome to find the problem is here.

brol1dev:
bump

PR #286, opened 2015-12-16
[PBA-2877] fixed more option row portrait and landscape switch bug

soapyigu:


brol1dev:
Cool! This is the flex way of doing things :)

PR #289, opened 2016-01-01
[PBA-2895] add error info for discover error screen

soapyigu:


sdk/react/discoveryPanel.js
```diff
@@ -198,6 +199,39 @@ var DiscoveryPanel = React.createClass({
         </Text>
       </View>);
   },
+
+  renderError: function() {
+    var errorTitleText = "SOMETHING NOT RIGHT! THERE SHOULD BE VIDEOS HERE.";
+    var errorContentText = "(Try Clicking The Discover Button Again On Reload Your Page)";
+    if (this.props.width < this.props.height) {
+      errorTitleText = "SOMETHING NOT RIGHT! THERE SHOULD\n BE VIDEOS HERE.";
```
ghost: can you use the Localization files?

ghost: Nevermind, I see

lental: This isn't the best way to do word wrapping.  there must be better ways to do this.  how will this work in smaller views?

sdk/react/discoveryPanel.js
```diff
@@ -198,6 +224,40 @@ var DiscoveryPanel = React.createClass({
...
+    var errorTitle = Utils.localizedString(this.props.locale, errorTitleText, this.props.localizedString);
+    var errorContent = Utils.localizedString(this.props.locale, errorContentText, this.props.localizedString);
+    var warningIcon = this.props.config.icons.warning.fontString;
+    if (this._isDicoveryError()) {
+      return (
+        <View style={[styles.panelErrorPanel, errorFlexDirectionStyle]}>
+          <View style={styles.panelErrorInfo}>
+            <Text style={styles.panelErrorTitleText}>
+              {errorTitle + "\n"}
+              <Text style={styles.panelErrorContentText}>
```
lental: this is fine, but I would have liked if the two <text> dom elements were separate, and not embedded in each other.  that way you would also be able to get rid of the \n, i think

lental:
sorry I didn't get to this earlier - You will need to fix merge conflicts.  You can also consider fixing what I mentioned in my comment above

sdk/react/discoveryPanel.js
```diff
@@ -119,6 +115,32 @@ var DiscoveryPanel = React.createClass({
...
       <Animated.View style={[styles.panel, animationStyle]}>
         {this.renderHeader()}
+        {this.renderList()}
+        {this.renderError()}
+      </Animated.View>
+    );
+  },
+
+  _isDicoveryError: function() {
+    return this.props.dataSource == null || this.props.dataSource.length == 0;
```
brol1dev: pls use === instead of ==

sdk/react/discoveryPanel.js
```diff
@@ -119,6 +115,32 @@ var DiscoveryPanel = React.createClass({
     return (
       <Animated.View style={[styles.panel, animationStyle]}>
         {this.renderHeader()}
+        {this.renderList()}
+        {this.renderError()}
+      </Animated.View>
+    );
+  },
+
+  _isDicoveryError: function() {
```
brol1dev: you missed an 's' in discovery :P, right now it is 'Dicovery'

sdk/react/discoveryPanel.js
```diff
@@ -119,6 +115,32 @@ var DiscoveryPanel = React.createClass({
...
+  },
+
+  renderList: function() {
+    var panelHeight = this.props.height - 40;
+    var renderHorizontal = Utils.shouldShowLandscape(this.props.width, this.props.height);
+    if (this._isDicoveryError()) {
+      return (
+        <ResponsiveList
+          horizontal={false}
+          data={null}
```
brol1dev, line 135: To understand a bit more, why do you use null here? If discovery has an error, it means that props.dataSource is empty, right?

sdk/react/discoveryPanel.js
```diff
@@ -200,6 +222,40 @@ var DiscoveryPanel = React.createClass({
...
+    var errorContentText = "(Try Clicking The Discover Button Again On Reload Your Page)";
+    var errorFlexDirectionStyle = {flexDirection: "row"};
+
+    if (this.props.width < widthThreshold) {
+      errorFlexDirectionStyle = {flexDirection: "column"};
+    }
+
+    var errorTitle = Utils.localizedString(this.props.locale, errorTitleText, this.props.localizedString);
+    var errorContent = Utils.localizedString(this.props.locale, errorContentText, this.props.localizedString);
+    var warningIcon = this.props.config.icons.warning.fontString;
```
brol1dev: could you check for the warning icon first, please? because right now, until we update the skin config to a newer version, we'll get an error because the warning icon doesn't exist in our current branch of skin config.

brol1dev:
LGTM

PR #291, opened 2016-01-04
PBA-3034

dmitrykb:


sdk/android/app/build.gradle
```diff
@@ -10,6 +10,9 @@ android {
         targetSdkVersion 23
         versionCode 1
         versionName "1.0"
+        ndk {
```
lental, line 13: can we add a comment to this? specifying that this is only because react-native does not have x64

PR #297, opened 2016-01-08
[PBA-3031] fix react loading race condition.

Matthew-C:
also change the method name to make it more clear.

### DIFF
--- a/sdk/iOS/OoyalaSkinSDK/OOSkinViewController.m
+++ b/sdk/iOS/OoyalaSkinSDK/OOSkinViewController.m
@@ -390,11 +390,17 @@ static NSDictionary *kSkinCofig;
   return OO_SKIN_VERSION;
 }
 
-- (void)queryState {
+- (void)onUIReady {
+  if (_player.currentItem) {
+    // embedcode already set, send current item information
+    NSNotification *notification = [NSNotification notificationWithName:OOOoyalaPlayerCurrentItemChangedNotification object:nil];
+    [self bridgeErrorNotification:notification];
+  }
+
   if (_player.state == OOOoyalaPlayerStateError) {
     NSNotification *notification = [NSNotification notificationWithName:OOOoyalaPlayerErrorNotification object:nil];
     [self bridgeErrorNotification:notification];
-  } else {
+  } else if (_player.state != OOOoyalaPlayerStateReady) {
     NSNotification *notification = [NSNotification notificationWithName:OOOoyalaPlayerStateChangedNotification object:nil];
     [self bridgeStateChangedNotification:notification];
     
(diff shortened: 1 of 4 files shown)


sdk/iOS/OoyalaSkinSDK/OOSkinViewController.m
```diff
@@ -390,11 +390,17 @@ - (NSString *) version {
...
+    // embedcode already set, send current item information
+    NSNotification *notification = [NSNotification notificationWithName:OOOoyalaPlayerCurrentItemChangedNotification object:nil];
+    [self bridgeErrorNotification:notification];
+  }
+
   if (_player.state == OOOoyalaPlayerStateError) {
     NSNotification *notification = [NSNotification notificationWithName:OOOoyalaPlayerErrorNotification object:nil];
     [self bridgeErrorNotification:notification];
-  } else {
+  } else if (_player.state != OOOoyalaPlayerStateReady) {
```
brol1dev: In JS, using the logical operator == or != sometimes may return an unexpected value, because it will try to do type conversion if the types are different. (ref: http://www.c-point.com/javascript_tutorial/jsgrpComparison.htm). I think we should try to always use === or !==, because this is safer, and is the equivalent to what we use in other languages.

lental: what happens on the overflow case?  I think if Autoplay is enabled, and react is super slow, the video might be in a playing state 

whoops, misread the condition.

If player state is ready... then it will have done the correct things in _player.currentItem above

Matthew-C: @rcvrgs this is obj-c, there's no such thing as ===. This is an enum comparison. I think I should change it into playable state(playing/paused etc)

brol1dev: LOL, my super bad, assumed this was JS xD

brol1dev:
I ran the app with this branch, using the query param: 'dev=false' and it doesn't fix it yet :( I tested with that param, as when we build jsbundle, it uses dev=false by default.

Should we be concerned with it right now? Maybe we could test a build and see if it works.

Matthew-C:
I saw play button appears but the location is off. It seems to me the state
is ok but configs are missing. I doubt there's racing/bugs in react code
here.

I generated the bundle file and put it into sample app with dev=false and I
don't see any problem.

Since our real scenario is to load bundle file locally I would say let's
not to waste time on this for now.

Thanks
Matthew

brol1dev:
Agreed @Matthew-C. I'll proceed and merge.

Thanks man!

PR #301, opened 2016-01-11
[PBA-2811] Pause video when view will disappear

brol1dev:
What do you think about this one? Added a flag that tells if the view will disappear, as the method is called when we make the back swipe gesture found in the PBA-2811 bug.

I decided to pause the video when we're in this animation, and resume the video if we get back to the playback view.

### DIFF
--- a/sdk/iOS/OoyalaSkinSDK/OOSkinViewController.m
+++ b/sdk/iOS/OoyalaSkinSDK/OOSkinViewController.m
@@ -41,6 +41,7 @@
 @property (nonatomic) OOUpNextManager *upNextManager;
 @property (nonatomic) NSDictionary *skinConfig;
 @property (nonatomic) BOOL isFullscreen;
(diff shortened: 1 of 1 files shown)


lental:
Better yet, can you find a way to straight remove the gesture?  it might make things easier to click :X

brol1dev:
@lental Ok, just pushed a new commit with the gesture disabled.

lental:
oh sorry, I don't mean add it to the SDK... A customer may still want to use it.. I think this issue is only isolated to the Sample App, correct?

brol1dev:
Yep, the problem is for the sample app, specifically to using the navigation bar. If we only want to fix for the sample app, then I'll close this PR and create a new one for the ios-sample-app.

lental:
Yeah, I think that would be fair for now.  I dont think we should dictate how the superviews and stuff deal with transitions, and that should be up to the application to figure out.

PR #303, opened 2016-01-11
water mark(control bar and on video) and updated bundle file.

wunnath:


sdk/react/widgets/videoWaterMark.js
```diff
@@ -0,0 +1,89 @@
...
+
+
+
+var logo = React.createClass({
+  propTypes: {
+    frameWidth: React.PropTypes.number,
+    frameHeight: React.PropTypes.number,
+    buttonWidth: React.PropTypes.number,
+    buttonHeight: React.PropTypes.number,
+    platform: React.PropTypes.string,
```
brol1dev: I didn't find any line that uses the platform in this file. It's okay if we remove it then.

sdk/react/widgets/videoWaterMark.js
```diff
@@ -0,0 +1,89 @@
...
+      height: new Animated.Value(0),
+    };
+  },
+
+ componentDidUpdate: function(prevProps, prevState) {
+    if(prevProps.width != this.props.width && this.props.isShow) {
+      this.state.height.setValue(ResponsiveDesignManager.makeResponsiveMultiplier(this.props.width, UI_SIZES.CONTROLBAR_HEIGHT));
+    }
+    if(prevProps.isShow != this.props.isShow ) {
+      this.state.opacity.setValue(this.props.isShow? 1 : 1);
```
brol1dev: If this line has the right logic, then it's the same if we just do setValue(1);

brol1dev:
LGTM. But you'll also need to create PR in skin-config to have the icons.

PR #304, opened 2016-01-13
Integrated Closed Caption Native component into UI

sveeravelli:


sdk/react/index.android.js
```diff
@@ -52,9 +52,9 @@ var OoyalaSkin = React.createClass({
       upNextDismissed: false,
       showPlayButton: true,
       // things which default to null and thus don't have to be stated:
-      selectedLanguage: null,
-      availableClosedCaptionsLanguages: null,
-      captionJSON: null,
+      // selectedLanguage: null,
```
lental, line 55: why are these commented?

sdk/android/sdk/src/main/java/com/ooyala/android/ooyalaskinsdk/OoyalaSkinLayoutController.java
```diff
@@ -261,6 +289,21 @@ private WritableMap getDiscovery() {
       return discoveryresults;
   }
 
+  private void getClosedCaptionView() {
```
lental: misnamed. this is more like "updateAvailableClosedCaptionsLanguages.

Also, how does this even work??? when do you even pass the languages up??

sdk/react/index.android.js
sveeravelli, line 55: It is for matching index.android.js to index.ios.js

PR #306, opened 2016-01-14
PBA-3106 - Fetch changes in android.js.bundle on the fly

dmitrykb:


lental:
good stuff, but can you add this to the app in Android-Sample-Apps?  you can do it in the 'dev' branch

PR #309, opened 2016-01-15
[PBA-2479] Make Share ActivityView work on fullscreen

brol1dev:
I pass SkinViewController as a dependency to OOActivityView, so that it is used as the presenting view controller for the UIActivityViewController. This way it is always shown, even on fullscreen.


brol1dev:
you're right. Didn't consider that scenario. Let me think in another way.

brol1dev:
I added a new logical functionality. Please review.

brol1dev:
Ok, thanks for that feedback. I renamed the method :)

PR #313, opened 2016-01-20
[PBA-2546] Check url stays empty if it is empty in react

brol1dev:
In our react code, we may send an empty link to be shared, because it isn't populated for that video content asset. In Obj-C the [RCTConvert NSURL:] was returning a non empty URL, pointing to an .app file that doesn't have to be shared at all.

This code makes sure that if react gives an empty url, it stays empty.

### DIFF
--- a/sdk/iOS/OoyalaSkinSDK/Social/OOActivityView.m
+++ b/sdk/iOS/OoyalaSkinSDK/Social/OOActivityView.m
@@ -28,8 +28,9 @@ RCT_EXPORT_METHOD(show:(NSDictionary *)options) {
     [items addObject:text];
   }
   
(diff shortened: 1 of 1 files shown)


lental:
I trust you tested a number of different scenarios (no url, with url, maybe consider invalid url).  If you can think of it in ~30 minutes or so, can you write a method and couple tests for the RCTConvert String -> RCTConvertURL steps you made?

brol1dev:
It's kinda hard to test that, as it involves RCT classes, I haven't been able to include those correctly in the Xcode test suite.

But i'm adding a new commit that makes the method to get the url more secure, as it ignores invalid urls too.

PR #318, opened 2016-01-27
PBA-2905 Implement Countdown Spinner for Android React Component

dmitrykb:


sdk/react/discoveryPanel.js
```diff
@@ -55,6 +58,20 @@ var DiscoveryPanel = React.createClass({
     };
   },
 
+  onTimerCompleted: function(e: Event) {
```
lental, line 65: add comments, saying this is only for android for now

sdk/android/ooyalaSkinSDK/src/main/java/com/ooyala/android/ooyalaskinsdk/CountdownView.java
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+
+    private void initPaints() {
+        textPaint = new TextPaint();
+        textPaint.setColor(textColor);
+        textPaint.setTextSize(textSize);
+        textPaint.setAntiAlias(true);
+
+
```
lental, line 51: whitespace

brol1dev:
It would be super nice if we only have one countdownTimer.js and not one for iOS and another one for Android. Do you wanna give a try in adding your countdownTimerAndroid.js logic into countdownTimer.js? React team does this for some of their components, take a look into Switch.js (https://github.com/facebook/react-native/blob/master/Libraries/Components/Switch/Switch.js) and you'll see what I'm talking about :)

@lental what do you think?

lental:
Talked with @dmitrykcom, unfortunately it's not easy to do that.

iOS utilizes some really sweet code block/method passing (see onTimerCompleted) that is unavailable in Android.  Android needs to use a different code path, and as such has a different set of props being passed.  There can be changes done to make the things more similar, but it cannot have the same interface at the end

dmitrykb:
Thanks Eric. That was definitely in my plans but I wasn't sure what is the best way to do it.
The problem is that I couldn't find the proper way how to handle callback from CountdownViewAndroid.
for ios component we use

```
onPress={this.onStatusPressed}
onTimerCompleted={() => this.onRowSelected(item)}
```

and then `RCTBubblingEventBlock` on the ios side.

Theres no such thing for android  and `discoveryPanel.js` had to register 

```
componentWillMount: function(e: Event) {
    DeviceEventEmitter.addListener('onTimerCompleted', this.onTimerCompleted)
}
```

I don't like it this way either but I'll dig into Switch view you suggested.

brol1dev:
Ok fair point, if it's hard to come with a quick solution, then we can leave it this way and think in improving later :)

PR #324, opened 2016-02-02
PBA-2905 upNext countdown

dmitrykb:


lental:
talked with @dmitrykcom, this will get the countdown showing, but the following things are not in the scope:
(1) clicking on upnext video does not play when you click
(2) the "x" dismiss button is not implemented
(3) the up next video does not play at the end of the video.

PR #328, opened 2016-02-05
UpNext Implementation 

sveeravelli:


sdk/android/ooyalaSkinSDK/src/main/java/com/ooyala/android/ooyalaskinsdk/BridgeMessageBuilder.java
```diff
@@ -18,8 +18,8 @@
  * Created by michael.len on 2/1/16.
  */
 class BridgeMessageBuilder {
-
-  public static WritableMap buildTimeChangedEvent (OoyalaPlayer player){
+  public static String nextVideoEmbedCode;
```
lental: why do you need to store this here? no storing anything in the BridgeMessageBuilder - it is entirely a static method class. no state in it

sdk/react/upNext.js
```diff
@@ -101,15 +101,16 @@ var UpNext = React.createClass({
       var upNextWidth = ResponsiveDesignManager.makeResponsiveMultiplier(this.props.width, thumbnailWidth, [0.8,1], [520]);
       var countdown = this.renderCountdownTimer();
       var upNextImage = (
-      <TouchableHighlight style={[styles.thumbnailContainer, {width: upNextWidth}]}
```
lental, line 104: not a fan of this.  the entire image should be clickable.  why did you change this?

sveeravelli:
Comments has been resolved.

sdk/react/upNext.js
lental, line 104: fixed with style change on touchable highlight below

PR #329, opened 2016-02-08
[PBA-3132] Fix bug of fullscreen inactive after rotating device twice

soapyigu:
The bug is caused by movieFullScreenView: usually it will set to current screen's bound. However, due to  the sensor inaccuracy of the device (I guess), the movieFullScreenView uses the previous landscape/portrait bound during portrait/landscape mode. By setting its bound to zero every time would force the view to reset if users tap fullscreen button; thus the problem shall not happen any more.


lental:
can you get this set up and show me after standup?

we should also check the Android side, make sure it's okay as well

PR #330, opened 2016-02-08
PBA-3117. Crash during ErrorScreen. Wrong function call.

dmitrykb:


sdk/react/index.android.js
```diff
@@ -110,11 +110,8 @@ var OoyalaSkin = React.createClass({
       case SCREEN_TYPES.END_SCREEN:   
         return OoyalaSkinCoreInstance.renderEndScreen();   
         break;
-      case SCREEN_TYPES.MOREOPTION_SCREEN:
-        return OoyalaSkinCoreInstance._renderMoreOptionScreen();
```
lental, line 114: any reason thsi was removed?

dmitrykb, line 114: Line 107 handles this case already and line 114 will never be executed. Also there's no method with name **_render**MoreOptionScreen

lental, line 114: jeeze, good catch.  thanks

PR #334, opened 2016-02-11
[PBA-2713] fix midroll and postroll bug

soapyigu:


lental:
I fixed this in a different way in my latest PR. please look

PR #337, opened 2016-02-12
[PBA-3318] check catalyst instance before emitting event.

Matthew-C:


sdk/android/ooyalaSkinSDK/src/main/java/com/ooyala/android/ooyalaskinsdk/OoyalaSkinLayoutController.java
```diff
@@ -434,4 +404,11 @@ private void requestDiscovery() {
             _player.getPcode(),
             ClientId.getId(_layout.getContext()), null, this);
   }
+
+  private void sendEvent(String eventName, ReadableMap params) {
```
lental, line 408: very nice.  this is much needed

PR #338, opened 2016-02-16
[PBA-3356] return empty notification if cc is not available, fixing NPE.

Matthew-C:


sdk/android/ooyalaSkinSDK/src/main/java/com/ooyala/android/ooyalaskinsdk/OoyalaSkinLayoutController.java
```diff
@@ -181,9 +181,15 @@ public void setFullscreenButtonShowing(boolean showing) {
 
   @ReactMethod
   public void onClosedCaptionUpdateRequested(ReadableMap parameters) {
-    final String languageName = parameters.hasKey("language") ? parameters.getString("language") : null;
-    double curTime = _player.getPlayheadTime() / 1000d;
-    WritableMap params = BridgeMessageBuilder.buildClosedCaptionUpdateParams(_player, languageName, curTime);
+    WritableMap params;
+    // ignore the request if cc is not available
+    if (_player == null || _player.getCurrentItem() == null || !_player.getCurrentItem().hasClosedCaptions()) {
+       params = Arguments.createMap();
```
lental, line 187: this is fine, but for the sake of discussion - What do you think about not even sending the callback response?  any reason you chose to send empty params instead?  what would an empty map do on the react side?

Matthew-C, line 187: not sending response is my first choice, but I found the last cc of previous video will stay there so I send an empty one to clear it. The alternative way is to reset the state when embedcode is set but if next video also have cc then user won't be able to see it and have to select it again.

lental, line 187: Any chance we can clear the state in a good way in the skin, without clearing the CC selection? It sounds like we _should_ be clearing the CC text when the skin gets the new video.

PR #347, opened 2016-02-23
[PBA-3260] Renamed the package to skin instead of ooyalaskinsdk

lental:


lental:
self-merging to get the build through

PR #348, opened 2016-02-23
fixed trimmed CC Language text and on/off switch

dmitrykb:
It does not mess up ios.

### DIFF
--- a/sdk/react/style/languageSelectionPanelStyles.json
+++ b/sdk/react/style/languageSelectionPanelStyles.json
@@ -16,6 +16,7 @@
     "alignItems": "center",
   },
   "panelTitle": {
+    "flex": 1,
     "padding": 10,
     "textAlign": "left",
     "margin": 10,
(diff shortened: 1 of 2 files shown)


sdk/react/style/languageSelectionPanelStyles.json
```diff
@@ -11,18 +11,20 @@
     "height": 88
   },
   "panelTitleRow": {
-    "height": 30,
+    "height": 40,
     "flexDirection": "row",
     "alignItems": "center",
   },
   "panelTitle": {
-  	"flex": 1,
```
lental, line 19: did you really have to remove the flex?

dmitrykb, line 19: yeah let's keep flex

lental:
I'm not really a huge fan of these changes - I kinda hate any pixel-related changes.  This should be easily flexbox'ed.  Can you confirm this will look okay on iOS?

PR #350, opened 2016-02-24
[PBA-3378] remove pause on clickad

Matthew-C:
this should be handled at ad plugin level. UI should not have that logic.


lental:
mmmmm fair point.  this may break some of our functionality, but i guess it's for the better

Matthew-C:
this can be easily added in managedadplayer, if necessary. But I don't think this is necessary. If learn more page is brought up, the app went to background and pause gets called anyway.

PR #357, opened 2016-02-26
Fixed playbutton clipping

dmitrykb:


lental:
not the biggest fan of this fix, but I understand the necessary code for getting the animation to be visible

PR #360, opened 2016-03-01
Jumping upNext animation

dmitrykb:


lental:
oh boy, good job.  0 px seemed to mess with Android's UI code

PR #361, opened 2016-03-01
[PBA-3327] Add left margin for countdown timer in Up Next

brol1dev:


brol1dev:
Verified UI in both iOS and Android.

wunnath:
LGTM

PR #362, opened 2016-03-01
[PBA-3328] Add description in up next view

brol1dev:
Verified UI for both iOS and Android.


lental:
looks really friggen smooth.  good job

PR #363, opened 2016-03-01
PBA-3239 color fix

wunnath:


sdk/react/style/controlBarStyles.json
```diff
@@ -38,7 +38,7 @@
     "paddingBottom":5,
     "flexDirection": "row",
     "alignSelf": "center",
-    "color": "white",
+    "color": "grey",
```
brol1dev: could you use the actual color code documented in the ticket?

wunnath: changed to hash codes provided in the ticket.

PR #364, opened 2016-03-01
[PBA-3326] iOS only - customize switch color

brol1dev:
This only works for iOS.

Currently in Android, the only way to change the color of the switch is through the Theme and Style APIs provided in native android. I don't know if we should go this way or just state in our documentation that it is the responsibility of the developer to change the color of a switch in their Android app through a theme. (ref: https://github.com/facebook/react-native/issues/2903)

### DIFF
--- a/sdk/react/widgets/ToggleSwitch.js
+++ b/sdk/react/widgets/ToggleSwitch.js
@@ -17,6 +17,7 @@ var Constants = require('../constants');
 
 var styles = require('../utils').getStyles(require('./style/ToggleSwitchStyles.json'));
 
+// Tint props only work for iOS.
 var ToggleSwitch = React.createClass({
   propTypes: {
     switchOn: React.PropTypes.bool,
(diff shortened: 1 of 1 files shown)


sdk/react/widgets/ToggleSwitch.js
```diff
@@ -24,7 +25,18 @@ var ToggleSwitch = React.createClass({
...
-    config: React.PropTypes.object
+    config: React.PropTypes.object,
+    onTintColor: React.PropTypes.string,
+    tintColor: React.PropTypes.string,
+    thumbTintColor: React.PropTypes.string,
+  },
+
+  getDefaultProps: function() {
+    return {
+      onTintColor: '#498DFC',
```
lental, line 36: where are these passed to the Switch?

brol1dev, line 36: Down below, in the render function.

lental, line 36: as in.. someone needs to create a new <ToggleSwitch>  Who does this, and how does that class derive those color values?

brol1dev, line 36: Oh, my bad. As a user of the ToggleSwitch, you can specify the color using those props. For example:

`<ToggleSwitch 
  onTintColor={'blue'}
  tintColor={'grey'}
  thumbTintColor={'white'}  />`

That would override the default props that I'm setting in the getDefaultProps function. Would you like me to add the color props where we are using the ToggleSwitch in the LanguageSelectionPanel.js?

lental, line 36: okay, so our SDK does not support the change of those values yet by the customer.  That's fine, I just wanted to understand that.

Thanks,
Mike

PR #365, opened 2016-03-01
code clean up and unmount of discovery panel fix

wunnath:
indexIos line is not used and causing a crash on android


lental:
oh boy, good catch

PR #375, opened 2016-03-08
localization from androidResource

dmitrykb:


lental:
you will need to update the skin-config submodule in ios-skin to support this

PR #380, opened 2016-03-10
Pba 3377

soapyigu:


sdk/react/adBar.js
```diff
@@ -88,11 +88,28 @@ var AdBar = React.createClass({
     return textString;
   },
 
+  getSkipTimerLabel: function() {
+    var skipTimerLabel = "";
```
Matthew-C: can we move this to utils? conversion between time number and string should be reusable.

PR #386, opened 2016-03-14
[PBA-3482] [PBA-3446] [PBA-3508] Fix Watermark to respect FlexBox

lental:
No more Absolute.  more Flex


lental:
mm shoudl remove somethin

PR #387, opened 2016-03-14
Fix Watermark to use FlexBox

lental:


lental:
bump, please review for the cut

PR #388, opened 2016-03-14
After selecting discovery no close button is visible and Video is getting stuck

sveeravelli:


lental:
conflicts btw, please resolve while we review

PR #390, opened 2016-03-15
skip add stylings in android and IOS

wunnath:


lental:
next time please  provide a better comment title

PR #392, opened 2016-03-16
Fix Bridge onPlayComplete to correctly put UI into pause state

lental:
DiscoveryEndScreen would show up and try to "Pause" the video, which would in fact play

We should seriously consider splitting the "playPause" into the two separate methods.


lental:
self merging for the release.  let me know if you have any particular concerns

PR #394, opened 2016-03-17
Up next Does not display any information on small player-fixed

wunnath:


wunnath:
added closedCaptionsLanguageChanged event creation for ooyalaskin(android) to this.

PR #397, opened 2016-03-23
Pba 3540

dmitrykb:


sdk/android/ooyalaSkinSDK/src/main/java/com/ooyala/android/skin/OoyalaSkinLayout.java
```diff
@@ -18,6 +21,16 @@ public void setFrameChangeCallback(FrameChangeCallback fcCallback){
       this.frameChangeCallback = fcCallback;
   }
 
+  @Override
+  protected void onLayout(boolean changed, int left, int top, int right, int bottom) {
+    if(initialHeight == -1) {
+      initialWidth = getWidth();
```
lental: I'm not convinced this is bug-free, but I don't have any good examples, stuff like animations, resizes, etc.

Maybe you can save the width and height when you actually go into fullscreen? maybe like a package-private method that stores that info in the layout

Or, it could be something that is saved in the LayoutController

PR #398, opened 2016-03-23
updated readme text to talk more generally about iOS AND Android

stephanieblotner:
Please also update the description of the repo to say iOS and Android


lental:
very cool.  thanks for updating!

PR #399, opened 2016-03-25
Fixed scrubber bar behind navigation bar after locking device.

dmitrykb:


sdk/android/ooyalaSkinSDK/src/main/java/com/ooyala/android/skin/OoyalaSkinLayoutController.java
```diff
@@ -420,6 +420,18 @@ public void onResume(Activity activity,
     if (_reactInstanceManager != null) {
       _reactInstanceManager.onResume( activity,defaultBackButtonImpl );
     }
+    // hide navigation and notification bars after lockscreen
+    // if video was in the fullscreen before screenlock
+    if(isFullscreen()) {
```
lental: maybe you can do setFullscreen(isFullscreen())? or, a helper function to make it so the flags aren't duplicated

PR #405, opened 2016-04-05
update skin-config to be tip of master, fix bottomOverlay to match

lental:


lental:
bump @rcvrgs , You can checkout https://github.com/ooyala/skin-config/commit/e794212c2df0a6ca297658ec874515c12feb8295 to see the context

note, default is that DVR is enabled, (the flag is disabled), and we should show the Scrubber during live by default

lental:
bump again.  after this, I may self-merge. the issue is not that intense

lental:
okay, firing away

PR #406, opened 2016-04-08
desired States implementation in IOS

wunnath:


wunnath:
added control bar icon scaling to center the icons in control bar.

PR #410, opened 2016-04-15
Fix view clipping in fullscreen mode on orientation change.

dmitrykb:


lental:
gotta put the ticket in the commit, please

PR #419, opened 2016-04-22
Pba 3384 ad overlay UI

Matthew-C:
@lental I tried to move overlay out of placeholder, the issue is:
1. overlay might cover play/pause button
2. the left/right area to the overlay will not respond to click.

therefore I decide to leave overlay inside the place holder


lental:
Sorry to keep going back and forth, but I'm honestly not convinced that the overlay needs to be in the placeholder.  The only thing causing the sides of the  view not to be clickable is the placeholder's `onTouchEnd={(event) => this.props.handlers.handleVideoTouch(event)}>`. We have CCs, Watermarks, and other bars that all should work the same.  

Re the overlay might cover the play-pause button, I don't see how putting it inside the placeholder helps. can you show me?

Matthew-C:
now ad overlay is out of placeholder.

lental:
okay, do  the empty left/right sides of the overlay respond to the click, or does it click-through to the ad?? I partially expected you to have to wrap the touchable-highlight with a view for that to work

Matthew-C:
now left and right area is clickable.

PR #428, opened 2016-05-03
[PBA-3469] need to specify frame size.

Matthew-C:
previously since there's no change video view might not get re-rendered on frame change. This should be ideally fixed by react native but the change works for now.


sdk/react/panels/videoView.js
```diff
@@ -325,13 +325,13 @@ var VideoView = React.createClass({
 
     return (
       <View
-        style={styles.container}>
```
lental, line 328: are you sure we don't want _any_ of styles.container?

PR #429, opened 2016-05-04
[PBA-3469] fix the issue with black magic.

Matthew-C:
the change in video view fixes a warning.


Matthew-C:
bump, can we CR/merge this so we can move this to QA?

PR #436, opened 2016-05-12
closed captions device and skin stylings support

wunnath:


wunnath:
@lental  @Matthew-C  please review.

PR #437, opened 2016-05-13
[PBA-3902] Listen for and use adError.

jonslenkooyala:


lental:
Nice, but long-term, we have to make sure that ad_pod_completed is here, and is the final decider when ad mode is gone

PR #439, opened 2016-05-16
new font update and watermark warning support.

wunnath:


wunnath:
@lental  @Matthew-C  please review this too.

lental:
can you create a new PR for this? this PR also has caption styling commits, which should not be part of this.  

making a new PR from this branch into dev should be enough

PR #442, opened 2016-05-24
[PBA-3985] remove zip file

Matthew-C:


lental:
do we need to add something to the gitignore?

Or will that be useless, because we are going to rename the zip file?

Matthew-C:
we already ignored OoyalaSkinSDK-*.zip so we don't need to add additional ignore

PR #452, opened 2016-06-01
removal of alice from dev docs and updating discovery and cc with new…

wunnath:
… font.


lental:
do you have to update android dev docs too?

wunnath:
not found any fonts info in android docs. android doc says - Put **fonts**,**index.android.jsbundle** and **skin.json** into **src/main/assets**.

PR #454, opened 2016-06-14
[PBA-3642] implement long press seek feature

soapyigu:


lental:
closing until we can get this to work more consistently

PR #455, opened 2016-06-16
FCC style support from IOS Device settings

wunnath:


wunnath:
@lental  resolved the conflicts. Please review it now.

sdk/react/ooyalaSkinPanelRenderer.js
```diff
@@ -112,6 +112,7 @@ OoyalaSkinPanelRenderer.prototype.renderVideoView = function() {
       captionStyles={{
         textSize: this.skin.state.ccTextSize,
         textColor: this.skin.state.ccTextColor,
+        fontName: this.skin.state.ccFontName,
```
lental, line 115: IS this used by Android?

wunnath, line 115: no. we are not using this for android. we are using it only for ios and it is not crashing android.

PR #466, opened 2016-07-19
[PBA-4033] add gradient and show/hide progress bar transition

soapyigu:


lental:
I look forward to see how this gets widget-ized

PR #471, opened 2016-07-23
[PBA-3931] fix Ooyala SkinSDK so player duration/playhead are passed …

lental:
…through the AdPodCompleteParams

Also changed the build.gradle to use project dir instead of user.dir

### DIFF
--- a/sdk/android/ooyalaSkinSDK/build.gradle
+++ b/sdk/android/ooyalaSkinSDK/build.gradle
@@ -33,9 +33,9 @@ repositories {
 }
 
 task copyLibs(type: Copy) {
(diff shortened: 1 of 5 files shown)


lental:
TODO: do the iOS side on Monday

lental:
bump, please review

lental:
another bump

brol1dev:
LGTM

brol1dev:
wait, i think this should be merged into dev not stable, right?

Matthew-C:
looks good to me too. Did we really want to merge to stable?

lental:
nope! good catch! switching it and merging

PR #477, opened 2016-08-16
[PBA-4426] implement swipe to seek

Matthew-C:


sdk/tvOS/OoyalaTVSkinSDK/OOTVGestureManager.m
```diff
@@ -89,4 +116,32 @@ - (void)togglePlay:(id)sender {
   [self.controller showProgressBar];
 }
 
+- (void)onSwipe:(id)sender {
+  if (sender == self.panGesture) {
+    CGPoint currentPoint = [self.panGesture translationInView:self.controller.view];
+    CGFloat seekScale = self.controller.player.duration / 1920.0;
```
brol1dev: Is 1920 a view width? Can it be retrieved differently instead of using a constant?

Matthew-C: yes, you are right, I should use view width. Actually we should allow users to configure the multiplier.

brol1dev: oh yea! that would be cool. Letting them choose from a range between a max limit (the view width) and a small limit, the lowest width we think it would look good, maybe a 1/3 of the view width or something like that.

brol1dev: oh, but I'm talking about the UI, this is swiping only. Oops!

Matthew-C:
please take a look.

sdk/tvOS/OoyalaTVSkinSDK/OOTVGestureManager.m
```diff
@@ -86,6 +86,13 @@ - (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer shouldBeRequi
   return NO;
 }
 
+- (BOOL)gestureRecognizerShouldBegin:(UIGestureRecognizer *)gestureRecognizer {
+  if ((gestureRecognizer == self.panGesture) && self.controller.player.isPlaying) {
+    return NO;
+  }
+  return YES;
```
lental, line 93: should this be "else if gesturerecognizer == self.panGesture"?

Matthew-C: we return yes by default, only return no for pan gesture when isPlaying is true.

lental: thanks, please add test guidelines to the ticket :)

PR #481, opened 2016-09-26
Old dev

lental:


lental:
During the release process, the dev branch ended up in a forced update that lost some of the progress on 'dev'.  This is merging what was lost back into 'dev' branch

brol1dev:
Oh, good thing you catch this. Did you make sure we're not missing any other commits?

lental:
based on the PRs, we should be good.

These are PRs with commits that were lost
PBA 4569 and 4600 : Updating react native to new version 0.33
[INTENTIONALLY INTO DEV] [PBA-4546] Prefer Desired State for determining what icons 
seek(start and complete) object support.

This PR was the last one that still exists in dev
[PBA-4426] implement swipe to seek 

This was the best I could do in trying to find lost code

PR #482, opened 2016-09-27
Added support for showAdMarquee and showControlBar from skin-config.json

akshaydch11:


sdk/react/panels/videoView.js
```diff
@@ -304,9 +304,8 @@ var VideoView = React.createClass({
        <ActivityIndicator
           style={loadingStyle}
           size="large"
-      />
-        );
-      }
+       />
+      );
```
pilievOoyala, line 308: do we not need the extra '}' anymore?

akshaydch11: No.. not needed ..  I changed some code in last pull request and missed to remove this one. So checking it in now:)

lental: good catch peter :)

but looking at the previous commit (https://github.com/ooyala/native-skin/commit/fef9a92f6a300796fba70acc2911811dcb121f2f) I can see we removed a layer of conditions, but didn't remove the last brace like akshay said

brol1dev:
LGTM

PR #484, opened 2016-09-28
support new parameter of seekInfo for skin and react.

wunnath:
seekEnd time support for seek complete notification.

### DIFF
--- a/sdk/react/ooyalaSkinBridgeListener.js
+++ b/sdk/react/ooyalaSkinBridgeListener.js
@@ -76,7 +76,7 @@ OoyalaSkinBridgeListener.prototype.onSeekStarted = function(e) {
 OoyalaSkinBridgeListener.prototype.onSeekComplete = function(e) {
   Log.log( "onSeekComplete");
   this.skin.setState({
-    playhead: e.playhead,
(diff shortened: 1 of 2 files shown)


sdk/react/ooyalaSkinBridgeListener.js
```diff
@@ -76,7 +76,7 @@ OoyalaSkinBridgeListener.prototype.onSeekStarted = function(e) {
 OoyalaSkinBridgeListener.prototype.onSeekComplete = function(e) {
   Log.log( "onSeekComplete");
   this.skin.setState({
-    playhead: e.playhead,
```
brol1dev, line 79: A question here. Do we have a plan about doing something with the playhead event? From this PR, I understand that the playhead passed to React is the seekStart position. Is it ok if we just discard it?

wunnath: re-named seekInfo variables for better understanding of them. at seekStart event, playhead should be at seekStart point and over seekComplete event, playhead should be the seekEnd point.

brol1dev:
I understand now. LGTM

PR #489, opened 2016-10-06
[PBA-4528] Add Closed Captions Scaling factor to Skin UI

lental:
Tested both Android and iOS with this code, looks fine


brol1dev:
LGTM!

PR #491, opened 2016-10-10
[PBA-4626] default controls to hidden initially, and support autohide…

lental:
… = false


sdk/react/ooyalaSkinCore.js
```diff
@@ -153,11 +153,15 @@ OoyalaSkinCore.prototype.handleVideoTouch = function(event) {
  * Hard reset lastPressedTime, either due to button press or otherwise
  */
 OoyalaSkinCore.prototype.handleControlsTouch = function() {
-  Log.verbose("handleControlsTouch - Time reset");
-  this.skin.setState({lastPressedTime: new Date().getTime()});
+  if (this.skin.props.controlBar.autoHide === true) {  
+    Log.verbose("handleVideoTouch - Time set");
```
brol1dev, line 157: Do we want these Logs?

brol1dev (approved):
LGTM, I just added one comment in a file.

sdk/react/ooyalaSkinCore.js
lental: logs are fine - after compiling the jsbundle, logs are removed.  futhermore, events that happen on-click are fine to log out to console, as it won't overflow the console output and gives good insight what happens

brol1dev: ok, that's all then :)

PR #492, opened 2016-10-11
	OoyalaSkinLayoutController is now observable and can listen to fullscreen events

akshaydch11:


sdk/android/ooyalaSkinSDK/src/main/java/com/ooyala/android/skin/OoyalaSkinLayoutController.java
```diff
@@ -315,6 +316,8 @@ public int getCurrentVolume() {
   @Override
   public void setFullscreen(boolean isFullscreen) {
     _layout.setFullscreen(isFullscreen);
+    setChanged();
+    notifyObservers();
```
lental: can you notify Observers with an OoyalaNotificationObject? see the OoyalaPlayer side:

`
  void sendNotification(String notificationName, Object data) {
    sendNotification(new OoyalaNotification(notificationName, data));
  }

  void sendNotification(OoyalaNotification notification) {
    setChanged();
    notifyObservers(notification);
  }
`

lental: I want to assume that we will eventually send multiple types of notifications eventually, and don't want to pigeonhole ourselves early

akshaydch11: Ok I will change it.

PR #494, opened 2016-10-12
Candidate fixes

lental:
Four independent fixes to various UI bugs.

These bugs popped up most likely due to the React 33 change


brol1dev:
LGTM

PR #500, opened 2016-11-01
[PBA-4485] Put Image inside main view, instead of acting as the parent

lental:
When the Promo Image is null, it causes the UI to not show up (All of image's children is hidden)


brol1dev:
Oh tricky bug, nice one :). LGTM!

PR #504, opened 2016-11-18
The jumping scrubber problem was caused by the playhead position gett…

pcaseooyala:
The jumping scrubber problem was caused by the playhead position getting updated to its old position at the end of the drag. The fix prevents updating the playhead position during the drag and sets the playhead position correctly at the end of the drag.

### DIFF
--- a/sdk/react/bottomOverlay.js
+++ b/sdk/react/bottomOverlay.js
@@ -33,7 +33,6 @@ var leftMargin = 20;
 var progressBarHeight = 3;
 var scrubberSize = 14;
 var scrubTouchableDistance = 45;
-var scrubPositionDelta = 0.2;
 var cuePointSize = 8;
 var BottomOverlay = React.createClass({
 
@@ -104,14 +103,12 @@ var BottomOverlay = React.createClass({
   },
 
 /* 
-If the difference between the old and new scrubber positions is greater than or equal to scrubPositionDelta, it is assumed
-that the scrubber is being dragged, and therefore, cachedPlayhead is set to -1 to prevent updating the scrubber position
-until the drag is done.
+If the playhead position has changed, reset the cachedPlayhead to -1 so that it is not used when rendering the scrubber
 */
   componentWillReceiveProps: function(nextProps) {
-    if (Math.abs(this.props.playhead - nextProps.playhead) >= scrubPositionDelta) { 
+    if (this.props.playhead != nextProps.playhead) { 
        this.setState({cachedPlayhead:-1.0}); 
-    }
+    } 
   },
 
   _renderProgressBarWidth: function() {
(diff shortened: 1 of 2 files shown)


sdk/react/bottomOverlay.js
```diff
@@ -100,6 +100,12 @@ var BottomOverlay = React.createClass({
     }
   },
 
+  componentWillReceiveProps: function(nextProps) {
+    if (Math.abs(this.props.playhead - nextProps.playhead) >= 0.2) {
+       this.setState({playhead:-1.0});
```
lental: can we rename playhead?  it makes the this.state.playhead and this.props.playhead confusing.

maybe something like cachedPlayhead, seeking Playhead, etc.

sdk/react/bottomOverlay.js
```diff
@@ -100,6 +100,12 @@ var BottomOverlay = React.createClass({
     }
   },
 
+  componentWillReceiveProps: function(nextProps) {
+    if (Math.abs(this.props.playhead - nextProps.playhead) >= 0.2) {
```
lental: can you add a comment here?  is this just an assumption that we send updates that are smaller than .2?  maybe instead you can do this.props.playhead != nextProps.playhead?

lental (changes requested):

lental:
Do we need to set an InitialState?  Can you look into that, and make sure it is initialized properly (it looks like it should be default -1, right?)?

sdk/react/bottomOverlay.js
pcaseooyala: We can't just do "!=" because then the cachedPlayhead will always be set to -1 and thus never be used.  I propose to define a var scrubPositionDelta = 0.2 and to add a comment before the function.

sdk/react/bottomOverlay.js
pcaseooyala: I vote for cachedPlayhead.

pcaseooyala:
Pushed the requested changes.  Please re-review.

pcaseooyala:
We need to have scrubPositionDelta because of timing issues.  Without it, if the scrubber is dragged, the scrubber gets rendered with the new position.  After a clock tick, componentWillReceiveProps gets called, but the props.playhead is still at the old position plus a clock tick.  cachedPlayhead gets set to -1, and the scrubber gets rendered with the props.playhead value which causes the scrubber to jump.  

Using scrubPositionDelta works because in componentWillReceiveProps, cachedPlayhead will only get set to -1 if the difference between the old and new playhead positions is more than just a clock tick (meaning the scrubber got dragged).  If componentWillReceiveProps gets called after a clock tick, cachedPlayhead will not get set to -1 (it will remain at the new, dragged position), and the scrubber will be rendered with the correct position.  Essentially, scrubPositionDelta is being used to detect a drag of the scrubber.

pcaseooyala:
Pushed the real fix.  ooyalaSkinBridgeListener.onSeekStarted caused problems by setting the playhead to seekstart instead of seekend.  With that fix, scrubPositionDelta is not needed, and bottomOverlay.componentWillReceiveProps can just compare props.playhead and nextProps.playhead for equality.

Also tested on iOS.

pcaseooyala:
Note: On iOS, for live streams, the scrubber jumps during seeks, but we will open a separate ticket for that.

Opened: https://jira.corp.ooyala.com/browse/PBA-4829

PR #505, opened 2016-11-21
Pba 4760

pcaseooyala:
Enabled user specification of scrubber color via skin.json.  Added a _customizeScrubber function to bottomOverlay.js that is called by _renderProgressScrubber.  _customizeScrubber defines the scrubber style with the scrubberHandleColor and scrubberHandleBorderColor from skin.json.


sdk/react/bottomOverlay.js
```diff
@@ -117,13 +117,20 @@ var BottomOverlay = React.createClass({
     var topOffset = this._renderTopOffset(scrubberSize);
     var leftOffset = this._renderLeftOffset(scrubberSize, percent, progressBarWidth);
     var positionStyle = {top:topOffset, left:leftOffset};
+    var scrubberStyle = this._customizeScrubber();
 
     return (
-      <View style={[styles.progressScrubber, positionStyle,{width:scrubberSize, height:scrubberSize}]}>
```
lental, line 122: can we delete progressScrubber style?

pcaseooyala: Do you mean "progressScrubber" style in bottomOverlayStyles.json?  I believe we can since this fix is defining the progressScrubber style inline.

lental: yeah the style in the .json file.  if we can delete it, we should.  alternatively, you can keep it (only remove the color parts), and define the coloring styles inline.  either way is good

sdk/react/bottomOverlay.js
```diff
@@ -117,13 +117,20 @@ var BottomOverlay = React.createClass({
...
+      <View style={[scrubberStyle, positionStyle,{width:scrubberSize, height:scrubberSize}]}>
       </View>
       );
   },
 
+  _customizeScrubber: function() {
+    var scrubberStyle = {flex: 0, position: "absolute", backgroundColor: this.props.config.controlBar.scrubberBar.scrubberHandleColor,
```
lental, line 129: In retrospect, this will break when people use an old "Skin.json". file.  

Is there any chance we can log an error  in react-native, if scrubberHandleColor and scrubberHandleBorderColor do not exist, we can raise an error, saying "controlBar.scrubberBar.scrubberHandleBorderColor not defined in your skin.json. update your skin.json file to the latest provided file, or add this to your skin.json"?  I don't know if we have any historical precedent for this, but this will happen often for people and we should have some warning

pcaseooyala: Another idea is to set default color values if scrubberHandleColor and scrubberHandleBorderColor are not defined.

pcaseooyala: I checked in a fix that logs error messages, but that will generate a lot of messages since it will log every time the scrubber is rendered.

PR #506, opened 2016-11-21
Pba 4760

pcaseooyala:
Added function bottomOverlay._customerScrubber to define the scrubber style.  It gets scrubberHandleColor and scrubberHandleBorderColor from skin.json, and it sets the other style values to default values.  Removed progressScrubber style from bottomOverlayStyles.json since it is no longer used.

If scrubberHandleColor and scrubberHandleBorderColor are not defined in skin.json, default values are used, and error messages are logged.


sdk/react/bottomOverlay.js
```diff
@@ -126,8 +127,18 @@ var BottomOverlay = React.createClass({
   },
 
   _customizeScrubber: function() {
-    var scrubberStyle = {flex: 0, position: "absolute", backgroundColor: this.props.config.controlBar.scrubberBar.scrubberHandleColor, 
-    borderRadius: 900, borderWidth: 1.5, borderColor: this.props.config.controlBar.scrubberBar.scrubberHandleBorderColor};
+    var scrubberHandleColor = this.props.config.controlBar.scrubberBar.scrubberHandleColor;
+    var scrubberHandleBorderColor = this.props.config.controlBar.scrubberBar.scrubberHandleBorderColor;
+    if (typeof scrubberHandleColor === "undefined") {
+      Log.error("this.props.config.controlBar.scrubberBar.scrubberHandleColor is not defined in your skin.json.  Please update your skin.json file to the latest provided file, or add this to your skin.json");
```
lental: customer will not know of this.props.config.  IMHO it would be more clear if we said controlBar.scrubberBar.scrubberHandleColor

pcaseooyala: I was wondering about that!  I'll make the changes.

lental:
lgtm. Thanks!

PR #507, opened 2016-11-29
Update Skin to support Bintray/Maven upload

lental:


brol1dev:
Cool! Does the react-native.aar also gets uploaded to bintray? How does that work?

lental:
react-native does not get uploaded to bintray, and like all local jar dependencies, they are required on the side.  see the example here: https://github.com/ooyala/android-sample-apps/compare/dev...maven

or: https://github.com/ooyala/android-sample-apps/blob/bb6e036ad06509bd676f5118f3bde4a351c5e9ab/ExoPlayerSampleApp/app/build.gradle#L89

brol1dev:
LGTM

PR #508, opened 2016-12-02
Pba 4780

pcaseooyala:
The problem was that the overlay's clickable area was larger than the visible overlay. Fixed the problem by making the visible overlay be the clickable object.

Fixed viewView._renderAdOverlay by changing changing the View that contains the image and the dismiss icon to a TouchableHighlight. Changed the outer TouchableHighlight to a View. Moved the onPress() to inside the new TouchableHighlight. Had to wrap the image and dismiss icon in a View because a TouchableHighlight can only have 1 child.


sdk/react/panels/videoView.js
```diff
@@ -277,24 +277,28 @@ var VideoView = React.createClass({
     var left = (this.props.width - width) / 2;
 
     return (
-      <TouchableHighlight
-        style={{height:height + 10, width:this.props.width}}
```
lental, line 281: any reason you removed the +10, and changed this.props.width to width?

lental:
going to merge this, so I can do a candidate build.  We may have to cherry-pick changes to this later

sdk/react/panels/videoView.js
pcaseooyala: No good reason.  I was experimenting and probably should have changed it back to original values.

PR #509, opened 2016-12-06
[PBA-4845] Inverse play/pause button in control bar widget

lental:
"If video is playing, I want to show the pause button to show what will happen when the button is pressed"
"If the video is paused, I want to show the play button"

primaryButton is only used in the play/pause button widget, if it is specified in the style.json

Delete unused adPlaybackScreen.  there is already one in the /panels/ folder.  I checked for references and could not find any


lental:
self merging for the candidate

PR #510, opened 2016-12-07
[PBA-4863] Avoid crashes with rgba() color settings in android skin

lental:


brol1dev:
LGTM! are we planning to support rgba?

lental:
yeah, react-native supports it fine, but java doesn't.  we'd have to catch rgba, and covert it into #FFFFFFFF

PR #512, opened 2016-12-09
[PBA-4894] Fix CachedPlayhead to only be set when actually seeking.

lental:
Touches are sometimes on other parts of the bottom overlay, and we set cachedPlayhead during those as well (touching duration, volume, etc)


brol1dev:
oh! good fix! LGTM

PR #514, opened 2016-12-14
Pba 4831

pcaseooyala:
In OOSkinPlayerObserver.m, added new methods getTotalDuration and getAdjustedPlayhead that use the seekable range to calculate and return total duration and adjusted playhead, respectively.  bridgeSeekStartedNotification and bridgeSeekCompletedNotification now call getTotalDuration to get useful duration values.  bridgeTimeChangedNotification now calls getTotalDuration and getAdjustedPlayhead.

### DIFF
--- a/sdk/iOS/OoyalaSkinSDK/OOSkinPlayerObserver.m
+++ b/sdk/iOS/OoyalaSkinSDK/OOSkinPlayerObserver.m
@@ -69,6 +69,8 @@
   }
 }
 
+
+// PBA-4831 Calculate total duration from the seekable range so that seek events have a correct value for duration
 - (NSNumber *) calculateTotalDuration {
     CMTimeRange seekableRange = _player.seekableTimeRange;
     Float64 duration;
--- a/sdk/iOS/OoyalaSkinSDK/OOSkinPlayerObserver.m
+++ b/sdk/iOS/OoyalaSkinSDK/OOSkinPlayerObserver.m
@@ -69,6 +69,17 @@
   }
 }
 
+- (NSNumber *) calculateTotalDuration {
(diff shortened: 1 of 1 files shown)


sdk/iOS/OoyalaSkinSDK/OOSkinPlayerObserver.m
```diff
@@ -69,13 +69,38 @@ - (void) addSelfAsObserverToPlayer:(OOOoyalaPlayer *)player {
   }
 }
 
+
+// PBA-4831 Return total duration calculated from the seekable range
+- (NSNumber *) getTotalDuration {
```
lental: I would prefer if these support methods take in a player as a parameter

pcaseooyala: No prob - can do

sdk/iOS/OoyalaSkinSDK/OOSkinPlayerObserver.m
```diff
@@ -100,7 +100,7 @@ - (void)bridgeSeekStartedNotification: (NSNotification *)notification {
   OOSeekInfo *seekInfo = seekInfoDictionaryObject[@"seekInfo"];
   NSNumber *seekStart = [NSNumber numberWithFloat:seekInfo.seekStart];
   NSNumber *seekEnd = [NSNumber numberWithFloat:seekInfo.seekEnd];
-  NSNumber *totalDuration = [self getTotalDuration];
+  NSNumber *totalDuration = [self getTotalDuration: (OOOoyalaPlayer *) _player];
```
lental: you should be able to use self.player here, and you shouldn't need to cast it

lental (approved):
LGTM

PR #515, opened 2017-01-10
Pba 4830

pcaseooyala:
Added support for specifying accentColor in skin.json.  Accent color is already supported on the web app.  It is used to specify color for the toggle thumb (iOS only - not supported on Android), volume control and buttons.  The accent color value will override a hard-coded value.


sdk/react/panels/languageSelectionPanel.js
```diff
@@ -147,8 +147,24 @@ var LanguageSelectionPanel = React.createClass({
     );
   },
 
+  getThumbTintColor: function() {
```
lental: hey is this used?

sdk/react/progressBar.js
```diff
@@ -27,11 +27,11 @@ var ProgressBar = React.createClass({
     var unbufferedPercent = 1 - playedPercent - bufferedPercent;
 
     if (this.props.ad) {
-      var playedColor = this.props.config.controlBar.adScrubberBar.playedColor;
+      var playedColor = this.props.config.controlBar.adScrubberBar.playedColor ? this.props.config.controlBar.adScrubberBar.playedColor : this.props.config.general.accentColor;
```
lental: this should have a condition if neither exist

sdk/react/controlBar.js
```diff
@@ -69,6 +69,14 @@ var ControlBar = React.createClass({
     }
   },
 
+  getVolumeControlColor: function() {
+    if (this.props.config.general.accentColor) {
+      return this.props.config.general.accentColor;
+    } else {
```
lental, line 80: can we add a condition here if neither exists?

lental (changes requested):

sdk/react/panels/languageSelectionPanel.js
pcaseooyala: No - will remove

sdk/react/controlBar.js
pcaseooyala: What do we do if neither exists?

sdk/react/progressBar.js
pcaseooyala: Ok, will set to a default value if neither exist

sdk/react/controlBar.js
pcaseooyala: Will set to a default value if neither exist

pcaseooyala:
Investigate:
 more options icons
 control bar icons
 play/pause

Should these be colored?

pcaseooyala:
Volume control bar - right now both sides of the bar are colored.  See if only the left side should be colored.

This is probably device-specific because the volume control bar is correctly colored on a simulated Nexus 6P.

sdk/react/controlBar.js
```diff
@@ -1,4 +1,4 @@
-/**
+  /**
```
lental: white space

pcaseooyala:
Suggestion: instead of changing the toggle thumb color, change the toggle background color.  The problem is that currently the toggle background is blue, and then changing the thumb color (from white) makes the resulting toggle look really bad.

Todo: ask Jeoffrey about this

pcaseooyala:
On iOS, the scrubber bar is not showing on the left side.  Investigate to see if we introduced a bug.

lental:
are there still changes to be added for this?

pcaseooyala:
Let me check again with Jeoffrey.  I didn't hear back from him earlier.

mistermendez:
hey @pcaseooyala let's sync up to review outstanding UI accent color changes

mistermendez:
@pcaseooyala here are some test links you can use to compare web accent color to native:
* https://goo.gl/IKbtQp (blue accent color)
* https://goo.gl/1N8lCc (red accent color)

mistermendez:
For web toggle switch, we change the switch background color, not the switch thumb color -

screen shot 2017-01-26 at 2 02 12 pm

mistermendez:
For web we do not apply accent color to control bar buttons, play/pause, or more options buttons.

The only time accent color is used for buttons is to indicate an active popover on desktop. Native does not have popovers so this won't apply to native-skin.

screen shot 2017-01-26 at 2 07 09 pm 2

sdk/react/controlBar.js
```diff
@@ -70,13 +70,15 @@ var ControlBar = React.createClass({
...
 
   getVolumeControlColor: function() {
-    if (this.props.config.general.accentColor) {
-      return this.props.config.general.accentColor;
-    } else if (this.props.config.controlBar.volumeControl.color) {
-      return this.props.config.controlBar.volumeControl.color;
+    if (!this.props.config.general.accentColor) {
+      if (!this.props.config.controlBar.volumeControl.color) {
+        Log.error("controlBar.volumeControl.color and general.accentColor are not defined in your skin.json.  Please update your skin.json file to the latest provided file, or add these to your skin.json");
+        return '#4389FF';
```
mistermendez, line 76: should we add these colors to constants.js?

sdk/react/progressBar.js
```diff
@@ -24,7 +24,7 @@ var ProgressBar = React.createClass({
   getAdScrubberBarPlayedColor: function() {
     if (!this.props.config.general.accentColor) {
       if (!this.props.config.controlBar.adScrubberBar.playedColor) {
-        Log.error("controlBar.adScrubberBar.playedColor and general.accentColor are not defined in your skin.json.  Please update your skin.json file to the latest provided file, or add this to your skin.json");
+        Log.error("controlBar.adScrubberBar.playedColor and general.accentColor are not defined in your skin.json.  Please update your skin.json file to the latest provided file, or add these to your skin.json");
         return '#FF3F80';
```
mistermendez, line 28: should we add these colors to constants.js?

sdk/react/progressBar.js
```diff
@@ -37,7 +37,7 @@ var ProgressBar = React.createClass({
   getScrubberBarPlayedColor: function() {
     if (!this.props.config.general.accentColor) {
       if (!this.props.config.controlBar.scrubberBar.playedColor) {
-        Log.error("controlBar.scrubberBar.playedColor and general.accentColor are not defined in your skin.json.  Please update your skin.json file to the latest provided file, or add this to your skin.json");
+        Log.error("controlBar.scrubberBar.playedColor and general.accentColor are not defined in your skin.json.  Please update your skin.json file to the latest provided file, or add these to your skin.json");
         return '#4389FF';
```
mistermendez, line 41: should we add these colors to constants.js?

pcaseooyala: This is an excellent idea, and it will be covered by a separate Jira issue.  [PBA-5368](https://jira.corp.ooyala.com/browse/PBA-5368)

PR #516, opened 2017-01-18
[PBA-5278] Fix Invalid React Start state by maintaining a queued even…

lental:
…t list before react is ready

This is not a public header, just internal

Also deleted code that was supposed to "emulate" all events before player is ready.  Ditch that and use a real system


brol1dev:
Verified! LGTM!

PR #517, opened 2017-01-25
Remove the Singleton functionality of the OOReact Bridge

lental:
Changed ReactBridge header to be all instance methods
Use instance properties instead of static RCTBridge and SkinViewController
add ViewController Event method for non-Controller classes to send events (instead of passing around an OOReactBridge)
OOUpNextManager doesn't need any ViewController functionality, so in that case I just pass the bridge. (I can be convinced of something better here)
VolumeManager really has nothing to do with message sending, move that into the ViewController

### DIFF
--- a/sdk/iOS/OoyalaSkinSDK/OOReactBridge.m
+++ b/sdk/iOS/OoyalaSkinSDK/OOReactBridge.m
@@ -23,8 +23,8 @@
 #import "OOConstant.h"
 
 @interface OOReactBridge()
-@property OOSkinViewController *controller;
-@property RCTBridge *rctBridge;
+@property (nonatomic, weak) OOSkinViewController *controller;
+@property (nonatomic, weak) RCTBridge *rctBridge;
 @end
 
 @implementation OOReactBridge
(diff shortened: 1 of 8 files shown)


sdk/iOS/OoyalaSkinSDK/OOReactBridge.m
```diff
@@ -22,12 +22,15 @@
 #import "OOUpNextManager.h"
 #import "OOConstant.h"
 
+@interface OOReactBridge()
+@property OOSkinViewController *controller;
```
brol1dev: do we want to add nonatomic attribute?

brol1dev (approved):
Super nice :) Code looks better!

sdk/iOS/OoyalaSkinSDK/OOSkinViewController.m
```diff
@@ -74,9 +74,13 @@ - (instancetype)initWithPlayer:(OOOoyalaPlayer *)player
     _isReactReady = NO;
 
     self.ooBridge = [OOReactBridge new];
+    //Passing self.ooBridge itself in the anonymous function counts as a self strong reference.
+    //I create a copy of the pointer to avoid that
+    OOReactBridge *newBridge = self.ooBridge;
```
brol1dev, line 79: Oh! Thanks for sharing this. Good to know :)

PR #518, opened 2017-02-02
[PBA-4352]  Support Volume tintColor  for iOS

lental:
React is **magical**! really, it just maps any input "color" prop to be a tintColor prop.

Also fixed an errant call to sendVolumeChangeEvent.  In a previous commit I removed that method implementation, had to remove it from the header as well


brol1dev:
Great stuff. So, for customers, they'll be able to change the color through code only, am I right? Do we want to use the `controlBar.volume.color` setting from skin.json?

lental:
yep! we already support volume color on Android's native component here: https://github.com/ooyala/native-skin/blob/stable/sdk/react/widgets/controlBarWidgets.js#L49

I just defined that color property to be remapped to the view's tintColor

brol1dev:
Ok, my bad. LGTM!

PR #519, opened 2017-02-07
Properly handle observers and React views for garbage collection

lental:
RootView needs to be unmounted
ReactInstanceManager needs to be destroyed
Volume Observer has to be unregistered.

This will happen as long as user correctly calls onDestroy on the destroy of the activity


brol1dev:
LGTM! Lets update our SkinSampleApp to show how to call onDestroy() and comment that it's required.

PR #522, opened 2017-02-17
Pba 5446

pcaseooyala:
This fixes the problem where messages are sometimes not getting localized. The problem was that in the skin, in ErrorScreen.js, the error message title was getting set to all caps before it was getting localized. Some of the keys in JSON files are all caps, and some are not. The fix was to set the title to all caps after being localized.


brol1dev:
cool! I can imagine how this was failing. LGTM!

PR #523, opened 2017-03-06
[PBA-5493] Show Discovery on end screen correctly

brol1dev:
We weren’t showing the discovery on the end screen with showUpNext = false and screenToShowOnEnd = “discovery”. This was a bug since we would only show discovery on end screen if we dismissed the upNext widget previously. Added a logic operation bypassing the upNextDismissed question if the showUpNext button was set to false, this way we would only care about screenToShowOnEnd.

I wanted to add testing to this change but I found out that with our current dependencies versions testing is broken. We will have to fix it first and then we should add more testing to our react native code to find this kind of errors.

### DIFF
--- a/sdk/react/ooyalaSkinCore.js
+++ b/sdk/react/ooyalaSkinCore.js
@@ -72,7 +72,7 @@ OoyalaSkinCore.prototype.handleMoreOptionsButtonPress = function(buttonName) {
       break;
     case BUTTON_NAMES.CLOSED_CAPTIONS:
       this.pushToOverlayStackAndMaybePause(OVERLAY_TYPES.CLOSEDCAPTIONS_SCREEN);
-      break; 
+      break;
     case BUTTON_NAMES.SHARE:
       this.ooyalaSkinPanelRenderer.renderSocialOptions();
       break;
@@ -133,7 +133,7 @@ OoyalaSkinCore.prototype.shouldShowLandscape = function() {
 };
 
 OoyalaSkinCore.prototype.shouldShowDiscoveryEndscreen = function() {
-  return (this.skin.state.upNextDismissed == true && this.skin.props.endScreen.screenToShowOnEnd == "discovery");
+  return (!this.skin.props.upNext.showUpNext || this.skin.state.upNextDismissed) && this.skin.props.endScreen.screenToShowOnEnd === "discovery";
 };
 
 /*
@@ -153,7 +153,7 @@ OoyalaSkinCore.prototype.handleVideoTouch = function(event) {
  * Hard reset lastPressedTime, either due to button press or otherwise
  */
 OoyalaSkinCore.prototype.handleControlsTouch = function() {
-  if (this.skin.props.controlBar.autoHide === true) {  
+  if (this.skin.props.controlBar.autoHide === true) {
     Log.verbose("handleVideoTouch - Time set");
     this.skin.setState({lastPressedTime: new Date()});
   } else {
--- a/sdk/react/ooyalaSkinCore.js
+++ b/sdk/react/ooyalaSkinCore.js
@@ -133,14 +133,18 @@ OoyalaSkinCore.prototype.shouldShowLandscape = function() {
 };
 
 OoyalaSkinCore.prototype.shouldShowDiscoveryEndscreen = function() {
+  var endScreenConfig = this.skin.props.endScreen || {};
+  var upNextConfig = this.skin.props.upNext || {};
+
   // Only care if discovery on endScreen should be shown
-  if (this.skin.props.endScreen.screenToShowOnEnd !== "discovery") {
+  if (endScreenConfig.screenToShowOnEnd !== "discovery") {
     return false;
   }
 
   // player didn't show upNext so show discovery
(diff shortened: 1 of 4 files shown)


brol1dev:
Remember we need two approvals before merging in the changes.

akshaydch11:
Looks good to me 👍

sdk/react/ooyalaSkinCore.js
```diff
@@ -133,7 +133,7 @@ OoyalaSkinCore.prototype.shouldShowLandscape = function() {
 };
 
 OoyalaSkinCore.prototype.shouldShowDiscoveryEndscreen = function() {
-  return (this.skin.state.upNextDismissed == true && this.skin.props.endScreen.screenToShowOnEnd == "discovery");
+  return (!this.skin.props.upNext.showUpNext || this.skin.state.upNextDismissed) && this.skin.props.endScreen.screenToShowOnEnd === "discovery";
```
aeng7: Should we split up this line into multiple lines for readability?

brol1dev: sure

brol1dev:
Updated code, @aeng7. I'm very used to use as little code as possible, sorry about that :P

brol1dev:
Oh wait, i'm going to change something else, don't merge yet!

brol1dev:
@akshaydch11, @aeng, review once more please, since I added some more changes in other impacted parts.

akshaydch11:
Looks Good 👍

sdk/react/ooyalaSkinCore.js
```diff
@@ -133,7 +133,24 @@ OoyalaSkinCore.prototype.shouldShowLandscape = function() {
 };
 
 OoyalaSkinCore.prototype.shouldShowDiscoveryEndscreen = function() {
-  return (this.skin.state.upNextDismissed == true && this.skin.props.endScreen.screenToShowOnEnd == "discovery");
+  // Only care if discovery on endScreen should be shown
+  if (this.skin.props.endScreen.screenToShowOnEnd !== "discovery") {
```
jpfranco: Do we populate this config with default values if, for example, `endScreen` is not provided?  I'm not sure if this would be an issue on the PBA side, but on the html5 skin we started using a safe get function so that this wouldn't crash if the skin config was incomplete.

brol1dev: pushed a change for what you're saying. This is why we need testing in place. We'll have to work towards it little by little. Thanks!

brol1dev:
@jpfranco please review once more when you have a chance

brol1dev:
I still need a second approval. Friendly reminder @jpfranco, @dchauOoyala, @aeng7. Ask me anything you need to know. 

Thanks :)

brol1dev:
guys, if no one comments today after our sprint planning I will merge it myself with one approval only. This PR is required for the upcoming release.

sdk/react/panels/EndScreen.js
```diff
@@ -69,10 +69,11 @@ var EndScreen = React.createClass({
   },
 
   _renderDefaultScreen: function(progressBar, controlBar) {
-    var fullscreenPromoImage = (this.props.config.endScreen.mode == 'default');
+    var endScreenConfig = this.props.config.endScreen || {};
+    var fullscreenPromoImage = endScreenConfig === 'default';
```
dchauOoyala: I notice the interchanging of single `''` and double `""` quotes throughout. Is there a reason for this?

sdk/react/panels/EndScreen.js
```diff
@@ -69,10 +69,11 @@ var EndScreen = React.createClass({
   },
 
   _renderDefaultScreen: function(progressBar, controlBar) {
-    var fullscreenPromoImage = (this.props.config.endScreen.mode == 'default');
+    var endScreenConfig = this.props.config.endScreen || {};
```
dchauOoyala, line 72: Is `this.props.config.endScreen` sometimes `String` and sometimes an object?

brol1dev: it is always an object, and thx for catching the error `endScreenConfig === 'default'` :)

sdk/react/panels/EndScreen.js
brol1dev: We should be using single quotes, but we have double quotes all over the place. I'll change the ones I see in the files I'm updating currently.

brol1dev: Oh btw, single quotes only apply to JS. Java and ObjC must use double quotes.

brol1dev: omg, we don't even use this fullscreenPromoImage. I'm just deleting this line.

brol1dev:
done @dchauOoyala, please review once more or merge if it's ok :)

dchauOoyala:
In regards to the quotes, there is not a mandated practice of using one over the other. It comes down to preference, but I personally prefer double quotes over single to remain consistent across languages. Web code currently leans towards double quotes. However, consistency is key so whichever one we do decide to use here for PBA we should do so moving forward.

dchauOoyala:
LGTM

PR #525, opened 2017-03-23
[PBA-5500] Update to react-native@0.35.0

brol1dev:
Update react native version from 0.33.0 to 0.35.0. This should potentially fix an issue with a customer.

Updates may be needed in the sample apps, but this one is required to build OoyalaSkinSDK.jar (Android) and OoyalaSkinSDK.framework (iOS) with the new version of react native.


akshaydch11:
Looks good ... one question dont we need to update react in "sdk/android/skin/libs/" folder ?

brol1dev:
No need because of the copyLibs script we have in skin/build.gradle. Actually, I think we may want to ignore the react-native-*.aar files in the skin/libs directory from git because of it.

Or practically, ignore everything that we copy into libs from the vendor folder.

brol1dev:
Added a new commit ignoring the skin/libs/ folder as I said in my previous comment.

akshaydch11:
Ok sounds good

dchauOoyala:
LGTM (don't know what these files are though)

brol1dev:
@dchauOoyala, the one you may know about is package.json. The other ones are updating the Android dependencies and code to match react-native@0.35.0.

brol1dev:
@akshaydch11, @dchauOoyala, could you approve and merge? With this I'll create a special RC for the customer.

PR #526, opened 2017-03-23
[PBA-5504] Make fullscreen property public

brol1dev:
This change is missing testing, I would like to add it but it may take me a good amount of time to set it up, since I need to be checking state of views, etc.

I'm sending this because it's blocking our release, if you see the code is okay we can merge it and we will still pass a QA cycle to manually catch if there are any errors.

### DIFF
--- a/sdk/iOS/OoyalaSkinSDK/OOSkinViewController.h
+++ b/sdk/iOS/OoyalaSkinSDK/OOSkinViewController.h
@@ -22,6 +22,24 @@
 @property (nonatomic, readonly) NSString *version;
 @property (nonatomic, readonly) OOClosedCaptionsStyle *closedCaptionsDeviceStyle;
 
+/**
+ Programatically change the fullscreen mode of the player.
+ Use this only after setting the OOSkinViewController.view.frame. Using it before will result in unexpected results.
+ 
+ Here is an example on how to use this property:
+ @code
+ // assume the property skinController of type OOSkinViewController exists and is already instantiated.
+ // Also assume you have an IBOutlet to a self.videoView pointing to an existing view.
+ [self addChildViewController:self.skinController];
+ self.skinController.view.frame = self.videoView.bounds;
+ // this will begin the player in fullscreen mode
+ self.skinController.fullscreen = YES;
+ 
+ [self.skinController.player setEmbedCode:MY_EMBED_CODE];
+ @endcode
+ */
+@property (nonatomic, getter=isFullscreen) BOOL fullscreen;
+
 // notifications
 extern NSString *const OOSkinViewControllerFullscreenChangedNotification; /* Fires when player goes FullScreen  */
 
@@ -35,4 +53,5 @@ extern NSString *const OOSkinViewControllerFullscreenChangedNotification; /* Fir
 - (void)ccStyleChanged:(NSNotification *) notification;
 
 - (void)sendBridgeEventWithName:(NSString *)eventName body:(id)body;
+
 @end
--- a/sdk/iOS/OoyalaSkinSDK/OOSkinViewController.h
+++ b/sdk/iOS/OoyalaSkinSDK/OOSkinViewController.h
@@ -24,19 +24,6 @@
 
 /**
  Programatically change the fullscreen mode of the player.
- Use this only after setting the OOSkinViewController.view.frame. Using it before will result in unexpected results.
- 
- Here is an example on how to use this property:
- @code
- // assume the property skinController of type OOSkinViewController exists and is already instantiated.
- // Also assume you have an IBOutlet to a self.videoView pointing to an existing view.
- [self addChildViewController:self.skinController];
- self.skinController.view.frame = self.videoView.bounds;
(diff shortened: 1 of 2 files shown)


akshaydch11:
Looking good.

sdk/iOS/OoyalaSkinSDK/OOSkinViewController.h
```diff
@@ -22,6 +22,24 @@
...
 @property (nonatomic, readonly) OOClosedCaptionsStyle *closedCaptionsDeviceStyle;
 
+/**
+ Programatically change the fullscreen mode of the player.
+ Use this only after setting the OOSkinViewController.view.frame. Using it before will result in unexpected results.
+ 
+ Here is an example on how to use this property:
+ @code
+ // assume the property skinController of type OOSkinViewController exists and is already instantiated.
+ // Also assume you have an IBOutlet to a self.videoView pointing to an existing view.
```
aeng7: What happens when these assumptions are wrong?

brol1dev: OOSkinViewController should exist always to use the fullscreen property, since it is part of it.

About self.videoView, in the example is needed because we need to set the frame of the skinViewController.view to a know value by the user. In the example it is assuming that there exists a view to which we reference through self.videoView, let's say that it's of size 100x100, so the player view will be of this size and positioned where the self.videoView is. The user can decide to set the frame with the value they want, but they must know that it has to be visible in the screen.

If they don't set the value of the SkinViewController.view.frame before calling to fullscreen, they will see an unexpected UI for the player, since it will use the default value of the SkinViewController.view.frame when trying to go fullscreen, since fullscreen depends on SkinViewController.view.frame.

sdk/iOS/OoyalaSkinSDK/OOSkinViewController.m
```diff
@@ -243,18 +242,17 @@ - (void)dealloc {
   [self.player destroy];
 }
 
-@end
-
-@implementation OOSkinViewController(Internal)
-
-- (void)toggleFullscreen {
+- (void)setFullscreen:(BOOL)fullscreen {
+  if (fullscreen == _fullscreen) return;
```
aeng7, line 246: is it possible for _fullscreen to be out of sync with the video player?

sdk/iOS/OoyalaSkinSDK/OOSkinViewController.h
aeng7: Should we add some logic to handle this case? Like if the value of view.frame is not set before calling fullscreen, we ignore the fullscreen call?

sdk/iOS/OoyalaSkinSDK/OOSkinViewController.m
brol1dev: it could be, that's why I need to add testing :(

brol1dev: I could see this happening if the method was called several times in a row. Something weird may happen in that case since we're pausing and resuming the player a lot of times, that may make it out of sync.

sdk/iOS/OoyalaSkinSDK/OOSkinViewController.h
brol1dev: yep, you're right. I'm looking at the code and it seems we're doing something bad. Thanks for asking this questions since I was able to notice :)

sdk/iOS/OoyalaSkinSDK/OOSkinViewController.m
aeng7: Can we check against the video player isFullscreen instead? This way we can have one source of truth.

brol1dev: good point. In iOS the rendering and UI state of the video player is managed by ourselves, outside of the player logic. Apple only gives a way to render video, but they don't have a concept of fullscreen, they just give you a view and you do whatever you want with it. That's why we need to maintain a state ourselves.

sdk/iOS/OoyalaSkinSDK/OOSkinViewController.h
brol1dev: @aeng7, it looks like we have good values of the frames in the init. I tested calling `skinController.fullscreen = YES;` just after creating the skinController instance and it works.

brol1dev:
FYI, @aeng7, @akshaydch11,

This is working now. I want to merge this as it is now but I'd like to come back to the OOSkinViewController in general because we need to test most of its methods. In the case of the setFullscreen method, it depends on having a value for the property `parentView`, if it doesn't fullscreen would break, but if we don't provide a parentView it would break the whole OOSkinViewController logic/UI, not just fullscreen.

Please ignore testing for this one so we can release and I'll note we need to review OOSkinViewController.

aeng7 (approved):

akshaydch11:
LGTM 👍

PR #529, opened 2017-05-05
[PLAYER-831] Reimplement fullscreen

brol1dev:
I was able to add a new fullscreen implementation without getting rid of a lot of code. Fullscreen now uses WindowManager to fill the whole screen, the previous solution was to get the current parent size and fill the parent, that wasn't real fullscreen.

This solution should work fine for Android 19 and above, Android 17 and 18 may have unexpected behavior, Android 16 and below will have a bad experience because it won't fill the whole screen in most cases.

Sadly, I wasn't able to add unit test for `OoyalaSkinLayout` and I don't think I will be able to add them soon because it is hard to unit test views in Android. There is a lot to it involved. What I am going to try to do is to have a good integration test suite but that would come later on.

If you want to test this you will have to link it to a sample app.


sdk/android/skin/src/main/java/com/ooyala/android/skin/OoyalaSkinLayout.java
```diff
@@ -94,45 +101,71 @@ public int getViewHeight() {
   }
 
   public boolean isFullscreen() {
-    return isFullscreen;
+    return fullscreen;
```
jpfranco: Should this refer to `this.fullscreen` to avoid confusing it with the parameter?

sdk/android/skin/src/main/java/com/ooyala/android/skin/OoyalaSkinLayout.java
```diff
@@ -94,45 +101,71 @@ public int getViewHeight() {
...
               View.SYSTEM_UI_FLAG_LAYOUT_STABLE);
     }
   }
 
   /**
-   * Stretch OoyalaSkinLayout to dimensions of parent layout.
+   * Stretch OoyalaSkinLayout to dimensions of the display window.
    * Handle system UI visibility.
    */
   void setFullscreen(boolean fullscreen) {
```
jpfranco, line 139: Can this be called before `FrameChangeCallback` is executed? If so then we might need to check to see if `windowManager` has been set.

sdk/android/skin/src/main/java/com/ooyala/android/skin/OoyalaSkinLayout.java
brol1dev: this method doesn't receive a parameter, or am i understanding the question incorrectly?

sdk/android/skin/src/main/java/com/ooyala/android/skin/OoyalaSkinLayout.java
brol1dev: yes it can and it's valid. I think you mean is `windowManager` used before it is even created? The answer should be no, but i am not a 100% sure.

I am initalizing `windowManager` whenever you create an instance of `OoyalaSkinLayout`, so I don't see it failing.

Anyway, I will be extra safe and add a null checker in case `windowManager` is not initialized for some strange reason.

akshaydch11:
It looks good. 👍

sdk/android/skin/src/main/java/com/ooyala/android/skin/OoyalaSkinLayout.java
jpfranco: Yes what I mean is that in other parts of the code we use `this.fullscreen` when referring to the field of the class and just `fullscreen` when it's a parameter. My suggestion is just to use `this.fullscreen` here in order to be consistent.

brol1dev:
I will automerge considering that you both approved this. Let me know tomorrow if you think I need to change something for this PR.

PR #534, opened 2017-07-26
PLAYER-1089 shows control bar when accessibility its enabled

agarcia001:


akshaydch11:
Is this PR ready to review ? 
You also need to add AccessibilityInfo for android as it alwaysOn controlbar behavior different on both platform.  We don't want that. Please make sure you test changed on android too.

brol1dev:
@agarcia001 do we still need this PR?

agarcia001:
Compare this solution with the Óscar ticket if not need this PR closet it

brol1dev:
I'll close this ticket.

PR #535, opened 2017-08-02
PBA-5502 Issues with Closed Captions

orobles-ooyala:
- Wrap closed captions to the next line.
- Do not scale closed captions text size when rotating the device, instead
  use a fixed 16 size.
- Place the closed captions at the same level as the watermark, and extend
  the closed captions to the full width if there is no watermark.
- Allow the closed caption menu to resize according to the player width,
  use a ScrollBack as a fallback if the height is not enough to show all
  the buttons.

### DIFF
--- a/sdk/react/panels/languageSelectionPanel.js
+++ b/sdk/react/panels/languageSelectionPanel.js
@@ -8,6 +8,7 @@ import React, { Component } from 'react';
 import {
   Animated,
   ListView,
+  ScrollView,
   StyleSheet,
   SwitchIOS,
   Text,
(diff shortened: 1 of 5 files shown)


sdk/react/panels/style/languageSelectionPanelStyles.json
```diff
@@ -55,6 +55,10 @@
     "alignSelf": "stretch",
     "backgroundColor": "#585858"
   },
+  "selectionPreviewtextContainer": {
```
brol1dev: Follow camel case for all the words, it should be `selectionPreviewTextContainer`. Can it be simpler? e.g. `previewTextContainer`

sdk/react/panels/languageSelectionPanel.js
```diff
@@ -117,35 +154,29 @@ var LanguageSelectionPanel = React.createClass({
...
-          horizontal={renderHorizontal}
-          data={this.props.languages}
-          itemRender={this.renderItem}
-          width={this.props.width}
-          height={itemPanelHeight}
-          itemWidth={160}
-          itemHeight={88}>
-        </ResponsiveList>
-        {previewText}
+        <ScrollView>
```
brol1dev: I think the ScrollView shouldn't wrap the header so it is always visible. Although I can see why you did it, is it because on smaller screens there is not enough space?

What do you think about having the `renderHeader` outside of the `ScrollView`?

brol1dev (commented):
Looks great, just added some inline comments.

Question, what are the styles you are using to wrap the CCs to the next line if they are too long?

brol1dev:
By the way, I just noticed there is a conflict with the `languageSelectionPreview.js` file, can you merge the latest changes in dev into this branch and fix the issues? Thanks.

sdk/react/panels/style/languageSelectionPanelStyles.json
orobles-ooyala: That was a typo :( , I will change it to previewTextContainer

sdk/react/panels/languageSelectionPanel.js
orobles-ooyala: It make sense, I will do some test to check if everything is ok

orobles-ooyala:
**Question, what are the styles you are using to wrap the CCs to the next line if they are too long?**

That is the default behavior unless a "flexDirection":"row" is given

brol1dev:
LGTM

brol1dev (approved):

PR #536, opened 2017-08-09
[PBA-5408] 

IvanSakharovskii:
Developed class for support full screen mode with RecyclerView

### DIFF
--- a/sdk/android/skin/src/main/java/com/ooyala/android/skin/util/FullScreenHelper.java
+++ b/sdk/android/skin/src/main/java/com/ooyala/android/skin/util/FullScreenHelper.java
@@ -1,0 +1,178 @@
+package com.ooyala.android.skin.util;
+
+import android.animation.Animator;
+import android.animation.AnimatorListenerAdapter;
+import android.animation.AnimatorSet;
+import android.animation.ObjectAnimator;
+import android.graphics.Point;
+import android.graphics.Rect;
+import android.view.View;
+import android.view.animation.DecelerateInterpolator;
+import android.widget.FrameLayout;
+
+/**
+ * FullScreenHelper is a class developed to support a full screen mode with RecyclerView
+ */
+
+public class FullScreenHelper {
+	private static final int ANIMATION_DURATION = 300;
+
+	private FrameLayout expandedLayout;
+	private FrameLayout currentParentLayout;
+	private Animator currentAnimator;
+	private FrameLayout currentPlayerLayout;
+
+	public FullScreenHelper(FrameLayout expandedLayout) {
+		this.expandedLayout = expandedLayout;
+	}
+
+	public void collapsePlayerLayout() {
+		if (expandedLayout.getVisibility() != View.VISIBLE || currentAnimator != null) {
+			return;
+		}
+
+		final Point globalOffset = new Point();
+		final Rect startBounds = new Rect();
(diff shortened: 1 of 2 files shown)


sdk/android/skin/src/main/java/com/ooyala/android/skin/util/FullScreenHelper.java
```diff
@@ -0,0 +1,178 @@
...
+		if (expandedLayout.getVisibility() != View.INVISIBLE) {
+			return;
+		}
+
+		if (currentAnimator != null) {
+			currentAnimator.cancel();
+		}
+
+		currentPlayerLayout = playerLayout;
+		currentParentLayout = (FrameLayout) playerLayout.getParent();
```
brol1dev: What if the `playerLayout.getParent()` does not return a `FrameLayout`? Can you add a validation here? If we don't get a FrameLayout this will break the app.

sdk/android/skin/src/main/java/com/ooyala/android/skin/util/FullScreenHelper.java
```diff
@@ -0,0 +1,178 @@
...
+ * FullScreenHelper is a class developed to support a full screen mode with RecyclerView
+ */
+
+public class FullScreenHelper {
+	private static final int ANIMATION_DURATION = 300;
+
+	private FrameLayout expandedLayout;
+	private FrameLayout currentParentLayout;
+	private Animator currentAnimator;
+	private FrameLayout currentPlayerLayout;
```
akshaydch11: Will it be good if we use OoyalaSkinLayout instead of generic FrameLayout as in very first pull request ?

brol1dev: Agree

brol1dev:
Friendly reminder to close this PR and reopen it to merge into `candidate` branch.

brol1dev:
Can you rename the class to something more meaningful? It occurs to me something like `RecyclerViewFullscreenManager` since that is the only purpose of this class.

sdk/android/skin/src/main/java/com/ooyala/android/skin/util/FullScreenHelper.java
brol1dev: You could even throw an exception that we always expect a `FrameLayout` since that is the use case for us. Otherwise we will need to validate the rest of the ViewGroups that exist.

sdk/android/skin/src/main/java/com/ooyala/android/skin/util/FullScreenHelper.java
```diff
@@ -0,0 +1,178 @@
...
+
+	private boolean isExtendedHorizontally(final Rect startBounds, final Rect finalBounds) {
+		float finalAspectRatio = (float) finalBounds.width() / finalBounds.height();
+		float startAspectRatio = (float) startBounds.width() / startBounds.height();
+
+		return finalAspectRatio > startAspectRatio;
+	}
+
+	private void cancelAnimation() {
+		FrameLayout parent = (FrameLayout) currentPlayerLayout.getParent();
```
brol1dev: same case here about the validation of `FrameLayout`

brol1dev (changes requested):

PR #538, opened 2017-08-10
Bugfix/pba 5408

IvanSakharovskii:
Renamed screen helper class, changed class type from FrameLayout to OoyalaSkinLayout.


sdk/android/skin/src/main/java/com/ooyala/android/skin/util/RecyclerViewFullScreenManager.java
```diff
@@ -0,0 +1,188 @@
...
+			public void onAnimationCancel(Animator animation) {
+				cancelAnimation();
+			}
+		});
+		set.start();
+		currentAnimator = set;
+
+	}
+
+	public void expandPlayerLayout(OoyalaSkinLayout playerLayout) {
```
brol1dev: can you add the `throws RuntimeException` clause in the definition of this method?

brol1dev (commented):
One last comment, can you add documentation to the public methods? So that the user know what they do. An overall explanation is enough.

brol1dev:
I'll merge it like it is right now, we'll add documentation later on.

PR #539, opened 2017-08-10
[PLAYER-1357] iOS UI issues when accessibility is enabled

orobles-ooyala:
- Avoid videoView superview to group its children into a single
  selectable component.
- Dismiss the controller presented modally to pass user touches
  to subviews.


brol1dev:
@orobles-ooyala can you reopen this ticket to merge into candidate instead of dev? We use candidate when we are in the middle of a release. I'll close this ticket and review the new one.

Thanks

PR #541, opened 2017-08-10
PBA-5679

IvanSakharovskii:
Added checking position in move method, now call touchEnd if touch was outside the scrubberLayout


brol1dev:
LGTM!

PR #542, opened 2017-08-10
[PLAYER-1357] iOS UI issues when accessibility is enabled

orobles-ooyala:
- Avoid videoView superview to group its children into a single
  selectable component.
- Dismiss the controller presented modally to pass user touches
  to subviews.


brol1dev:
nice one! LGTM!

PR #545, opened 2017-08-10
Play pause

agarcia001:


brol1dev:
There are warnings appearing in the player because you are applying unsupported style properties to the parent `View` that you added.  A `View` doesn't accept `textAlign` and `color` properties. Please change this. 

Be careful about the styles that you apply to `View` and `Animated.Text`

brol1dev:
LGTM

PR #549, opened 2017-08-15
[PBA-5856] Check if VoiceOver is on

brol1dev:


brol1dev:
self merging this one since we need it to have a better build candidate today.

PR #550, opened 2017-08-24
[Accessibility] Add Accessibility state modules

brol1dev:
Added a module for both iOS and Android to know about the current state of the screen readers for each platform.

Ask me if you want to test it out but I checked it by adding some snippets to both index.ios.js and index.android.js.

From here we need to customize the UI depending on the state of a screen reader, but that will come in a different PR.

Note that when we update to the latest version of react native, we will need to remove the files and changes applied with this PR, since they will be part of react native. With this PR we are just adding the Accessibility APIs added in newer versions of react native.


akshaydch11:
Looks good 👍

PR #558, opened 2017-09-04
[PBA-5915] Comment out setAccessibilityFocus usage

brol1dev:
setAccessibilityFocus is not part of the code in react native 0.35.0. We need to come up with a way to have it available in this version.

Right now the work around is just to comment out wherever we used it.


brol1dev:
self merging to let QA continue testing.

PR #564, opened 2017-09-11
Update native skin for VR

IvanSakharovskii:
Update native skin functionallity for VR


sdk/android/skin/build.gradle
```diff
@@ -114,7 +114,7 @@ repositories {
 }
 
 task copyLibs(type: Copy) {
-    from new File(projectDir, '../../../vendor/Ooyala/OoyalaSDK-Android/OoyalaSDK.jar')
```
brol1dev, line 117: what happened to the `OoyalaSDK.jar`? Shouldn't OoyalaVRSDK depend on OoyalaSDK?

brol1dev (commented):
I feel i gotta see the changes android-sdk first because there's a lot of code that is being used here that I don't understand.

sdk/android/skin/build.gradle
IvanSakharovskii: I created a new JAR file from our SDK branches. After we will update dev branch in android SDK  we will return previous path.

PR #567, opened 2017-09-14
[PLAYER-1736] When screen reader is enabled, change playhead by 10 seconds

brol1dev:


brol1dev:
@akshaydch11, updated the code, please review again.

PR #604, opened 2017-11-29
[PLAYER-2243]

atopilskii:
Add set seek time to 0 if there is negative value

### DIFF
--- a/sdk/android/skin/src/main/java/com/ooyala/android/skin/OoyalaSkinLayoutController.java
+++ b/sdk/android/skin/src/main/java/com/ooyala/android/skin/OoyalaSkinLayoutController.java
@@ -419,7 +419,11 @@ public class OoyalaSkinLayoutController extends Observable implements LayoutCont
         sendEvent(CONTROLLER_KEY_PRESS_EVENT, null);
         break;
       case KeyEvent.KEYCODE_MEDIA_REWIND:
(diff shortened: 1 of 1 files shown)


sdk/android/skin/src/main/java/com/ooyala/android/skin/OoyalaSkinLayoutController.java
```diff
@@ -419,7 +419,11 @@ public boolean onKeyDown(int keyCode, KeyEvent event) {
         sendEvent(CONTROLLER_KEY_PRESS_EVENT, null);
         break;
       case KeyEvent.KEYCODE_MEDIA_REWIND:
-        _player.seek(_player.getPlayheadTime() - REWIND_STEP); // << -10sec
+        int timeAfterSeek = _player.getPlayheadTime() - REWIND_STEP;
+        if(timeAfterSeek < 0) {
+          timeAfterSeek = 0;
+        }
```
AlinaVoronkovaEpam: Use smth like this: `= Math.max(0, ..)` instead of `if(timeAfterSeek < 0) { ...`

AlinaVoronkovaEpam (changes requested):

AlinaVoronkovaEpam (approved):

IvanSakharovskii (approved):

PR #622, opened 2017-12-22
[PLAYER-2604] Update ooyala-slick-type.ttf to include cardboard icon

brol1dev:
Copied the font from the html5-skin repo: https://github.com/ooyala/html5-skin/blob/master/assets/fonts/ooyala-slick-type.ttf


aldigun:
LGTM.

PR #623, opened 2017-12-29
Dev epam

IvanSakharovskii:


brol1dev:
@IvanSakharovskii, how are you using the testId attributes? Why are they necessary?

IvanSakharovskii:
This changes is necessary for QA automation. We checked previous tests and they work fine. Now we are adding a new autotests for VR.

PR #627, opened 2018-01-15
[PLAYER-2639] OoyalaSDK: jar was changed on aar

AlinaVoronkovaEpam:


brol1dev:
Rollback the binaries (.jar, .aar) from the PR. They will be updated with the build scripts.

akshaydch11 (approved):

PR #628, opened 2018-01-18
[PLAYER-2157]

atopilskii:
Add catch error for geoblocking

### DIFF
--- a/sdk/react/utils.js
+++ b/sdk/react/utils.js
@@ -215,6 +215,9 @@ var Utils = {
       /* Unknown error - TODO add to language files */ 
       case 28:
         return "An unknown error occurred";
+      /* GeoBlocking access denied - TODO add to language files */
+      case 31:
+        return "Geo access denied";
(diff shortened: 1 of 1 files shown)


AlinaVoronkovaEpam (approved):

sdk/react/utils.js
```diff
@@ -215,6 +215,9 @@ var Utils = {
       /* Unknown error - TODO add to language files */ 
       case 28:
         return "An unknown error occurred";
+      /* GeoBlocking access denied - TODO add to language files */
+      case 31:
```
brol1dev: were are the cases 29 and 30?

atopilskii: I think there is an mistake. We have 31 errors in class OoyalaException (sdk), but here i see only 28 errors. I will add missed error codes.

pilievOoyala:
LGTM

brol1dev (approved):

PR #632, opened 2018-01-29
[PLAYER-1740] Add accessibility label to play icon in up next widget

brol1dev:


sdk/react/upNext.js
```diff
@@ -130,8 +130,13 @@ var UpNext = React.createClass({
           source={{uri: this.props.nextVideo.imageUrl}}
           style={styles.thumbnail} >
           <TouchableHighlight style={styles.thumbnail}
-            onPress={this.clickUpNext}>
-            <Text style={styles.countdownText}>{this.props.config.icons.play.fontString}</Text>
+            onPress={this.clickUpNext}
+            accesible={true}
```
pilievOoyala: accesible -> accessible

sdk/react/upNext.js
```diff
@@ -83,7 +83,7 @@ var UpNext = React.createClass({
   },
 
   _renderDismissButton: function() {
```
pilievOoyala, line 84: Where the docs at?!?!?!

sdk/react/upNext.js
```diff
@@ -130,8 +130,13 @@ var UpNext = React.createClass({
           source={{uri: this.props.nextVideo.imageUrl}}
           style={styles.thumbnail} >
           <TouchableHighlight style={styles.thumbnail}
-            onPress={this.clickUpNext}>
-            <Text style={styles.countdownText}>{this.props.config.icons.play.fontString}</Text>
+            onPress={this.clickUpNext}
+            accesible={true}
+            accessibilityLabel={"Play up next video"}>
```
akshaydch11: Can we add it as constant variable ? Something like "BUTTON_NAMES.DISMISS"

sdk/react/upNext.js
brol1dev: fair enough, but to be honest I think we need to work on docs on all the files in our JS code since we have none. Should I start on this one or do I wait and start on another PR just dedicated to docs?

sdk/react/upNext.js
brol1dev: Done

sdk/react/upNext.js
brol1dev: thanks

sdk/react/upNext.js
pilievOoyala: To minimize downtime, if you touch a function, look at it's jsdocs.  You don't have to do the whole file.  We'll chip away at it slowly.

pilievOoyala:
LGTM. We can address the documentation after this release.

PR #634, opened 2018-02-08
[PLAYER-1778] Fix layout for iOS platform

M0rtyMerr:
This fix looks like a crutch, but it's not.
Layout is actually ok - i can't reproduce this bug with android. But both platform use their own native element for volume slider.
MPVolumeView's positioning have some kind of bug (or something that i just can't understand).
During my investigation i wrapped VolumeView in a simple React Native View and tried to positioning it.  Wrapper View was ok (aligned just how it should be), but Volume View just go out from super view's borders. 
So, finally, bug is definitely not in layout. There is problem with MPVolumeView, which is always has 
"Y: -5" instead of "Y: 0"  (tested with iPad pro, ipod touch, iphone 6).
This fix resolve bug for these devices and has no effect with Android.

### DIFF
--- a/sdk/react/controlBar.js
+++ b/sdk/react/controlBar.js
@@ -150,7 +150,8 @@ var ControlBar = React.createClass({
         showVolume: this.state.showVolume,
         volume: this.props.volume,
         scrubberStyle: styles.volumeSlider,
-        volumeControlColor: this.getVolumeControlColor()
+        volumeControlColor: this.getVolumeControlColor(),
+        platform: this.props.platform
       },
       timeDuration: {
         onPress: this.props.live ? this.props.live.onGoLive : null,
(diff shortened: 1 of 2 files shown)


IvanSakharovskii (approved):

brol1dev (approved):
LGTM

brol1dev:
I just have a question @MortMerr, @IvanSakharovskii, how is this issue related to PLAYER-1788? I see that the ticket is a web issue.

IvanSakharovskii:
@rcvrgs This bug is related to Sample Apps. Sorry for error, bug number is 1778. Here's link: https://jira.corp.ooyala.com:8443/browse/PLAYER-1778

MarkEPAM (approved):

PR #672, opened 2018-04-24
Added localizations for undefined audio track language and no linguistic content

MarkEPAM:
Added localizations for undefined audio track language and no linguistic content.
Created constant object in AudioAndCCSelectionPanel.js for strings.

### DIFF
--- a/sdk/react/panels/AudioAndCCSelectionPanel.js
+++ b/sdk/react/panels/AudioAndCCSelectionPanel.js
@@ -22,6 +22,14 @@ var Utils = require('../utils');
 var styles = require('../utils').getStyles(require('./style/AudioAndCCSelectionPanel'));
 var ItemSelectionScrollView = require('./ItemSelectionScrollView');
 
+const stringConstants = {
+  undefinedLanguageTitle: "Undefined language",
+  noLinguisticContentTitle: "No linguistic content",
+  offButtonTitle: "Off",
+  audioHeaderViewSectionTitle: "Audio",
+  subtitlesHeaderViewSectionTitle: "Subtitles"
+}
+
 var AudioAndCCSelectionPanel = React.createClass({
   propTypes: {
     audioTracksTitles: React.PropTypes.array,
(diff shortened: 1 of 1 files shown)


brol1dev (approved):
LGTM

sdk/react/panels/AudioAndCCSelectionPanel.js
```diff
@@ -56,13 +64,20 @@ var AudioAndCCSelectionPanel = React.createClass({
   },
 
   onAudioTrackSelected: function(name) {
-    if (this.props.selectedAudioTrackTitle !== name) {
-      this.props.onSelectAudioTrack(name);
+    let localizedTitleForUndefinedLanguage = Utils.localizedString(this.props.config.locale, stringConstants.undefinedLanguageTitle, this.props.config.localizableStrings);
+    let localizedTitleForNoLinguisticContent = Utils.localizedString(this.props.config.locale, stringConstants.noLinguisticContentTitle, this.props.config.localizableStrings);
+    var originalName = name;
```
pilievOoyala, line 69: I think you have the naming of these two variable backwards. If we edit the name, it's no longer the original name. I would rename the param that's passed in to be 'originalName', and the new localized one to be 'name'.

pilievOoyala: Disregard the previous statement. I got the parameters mixed up.

sdk/react/panels/AudioAndCCSelectionPanel.js
```diff
@@ -56,13 +64,20 @@ var AudioAndCCSelectionPanel = React.createClass({
...
+
+    originalName = originalName.replace(localizedTitleForUndefinedLanguage, stringConstants.undefinedLanguageTitle);
+    originalName = originalName.replace(localizedTitleForNoLinguisticContent, stringConstants.noLinguisticContentTitle);
+
+    if (this.props.selectedAudioTrackTitle !== originalName) {
+      this.props.onSelectAudioTrack(originalName);
```
pilievOoyala, line 75: I cannot tell what the different is between onSelectAudioTrack and onAudioTrackSelected.  I can see what this function is doing but if I were to read these two function names someone else in the code, I would not know which one to use. Can you explain the difference and maybe we can name them better?

pilievOoyala:
LGTM

sdk/react/panels/AudioAndCCSelectionPanel.js
pilievOoyala: We can do this in another PR.

PR #675, opened 2018-05-01
[PLAYER-3389] Use Error messages defined in constants if found

brol1dev:
Related to https://git.corp.ooyala.com/projects/PBA/repos/ios-sdk/pull-requests/297/overview. Now that we receive the correct error message and code for device limit reached I'm defining some constants that define the message shown in the UI.


AlinaVoronkovaEpam (approved):

sdk/react/utils.js
```diff
@@ -129,84 +129,84 @@ var Utils = {
...
+      /* Invalid authorization token */
       case 10:
-        return "INVALID PLAYER TOKEN";
-      /* Device limit has been reached */    
+        return "Invalid Player Token";
+      /* Device limit has been reached */
       case 11:
-        return "Device limit has been reached";
-      /* Devuce binding failed */    
+        return "Authorization Error";
```
pilievOoyala, line 166: Do we not need the device limit error message anymore? Did it move to a different number? I just don't see it.

brol1dev: I removed it as required by Product (Steve, Nikhil) They want the message to be Authorization Error. We still send device limit error to the message bus is just different how we show the message in the SkinSDK.

pilievOoyala:
LGTM. @akshaydch11 Can you review as well?

AlinaVoronkovaEpam (approved):

akshaydch11 (approved):

PR #676, opened 2018-05-03
[PBA-6293] Added logic for hide home indicator on iPhone X

MarkEPAM:
- PR resolves problem with showed home indicator on iPhone X. 
- Were added functions for view controllers. 
- PR without unit tests, because they are not required.

### DIFF
--- a/sdk/iOS/OoyalaSkinSDK/ViewControllers/OOSkinFullScreenViewController.m
+++ b/sdk/iOS/OoyalaSkinSDK/ViewControllers/OOSkinFullScreenViewController.m
@@ -38,6 +38,10 @@
 
 #pragma mark - Override view controller functions
 
+- (BOOL)prefersHomeIndicatorAutoHidden {
+  return YES;
+}
+
 - (UIInterfaceOrientationMask)supportedInterfaceOrientations {
   return _supportedInterfaceOrientations;
 }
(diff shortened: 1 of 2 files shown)


brol1dev (approved):
LGTM

sdk/iOS/OoyalaSkinSDK/ViewControllers/OOSkinFullScreenViewController.m
```diff
@@ -38,6 +38,10 @@ - (instancetype)init {
 
 #pragma mark - Override view controller functions
 
+- (BOOL)prefersHomeIndicatorAutoHidden {
+  return YES;
```
pilievOoyala, line 42: Do we need to do some kind of check to see if this is an iPhoneX? If not, will this effect older phones like the 8.  What about future phones?

sdk/iOS/OoyalaSkinSDK/ViewControllers/OOSkinViewController.m
```diff
@@ -201,6 +201,10 @@ - (void)dealloc {
 
 #pragma mark - Override view controller functions
 
+- (BOOL)prefersHomeIndicatorAutoHidden {
+  return YES;
```
pilievOoyala, line 205: Same thing here.

sdk/iOS/OoyalaSkinSDK/ViewControllers/OOSkinFullScreenViewController.m
IvanSakharovskii: Only iPhone X have home indicator. For previous versions this function won't be called.

brol1dev: it will only be called for iPhone X @pilievOoyala

pilievOoyala:
Looks good. Thank you for clarifying.

PR #679, opened 2018-05-24
[PLAYER-3521] Safely send error.userInfo

brol1dev:
NSDictionary does not accept nil values when using the definition syntax `@{key: value}`.


pilievOoyala:
LGTM

PR #681, opened 2018-05-28
[PLAYER-3149] - [Apple TV] Closed captions are not shown when selected in TV settings.

ipadilla4:
Calling SDK method to avoid CC duplicates when using Skin SDK.

https://jira.corp.ooyala.com/browse/PLAYER-3149


brol1dev:
is there a case in which the CC from the HLS Playlist (m3u8) are shown using the SkinSDK? Or does the Skin only support CC from content tree?

ipadilla4:
My understanding is that Skin only supports CC from content tree to avoid duplicates.

brol1dev (approved):
LGTM, read comments added in our internal repos.

akshaydch11:
LGTM

PR #687, opened 2018-06-14
Update react-native to version 0.55.4

IvanSakharovskii:
Updated libraries for iOS and android, refactored JS code with new react-native standards.

PR #688, opened 2018-06-14
[PLAYER-3511] Fix bug with End screen in react native

atopilskii:
Added showing load spinner after press Repeat button
Add checking of current screen to prevent showing Play button instead of Replay button on End screen 
There is no test because I cannot cover it by test

### DIFF
--- a/sdk/react/ooyalaSkinBridgeListener.js
+++ b/sdk/react/ooyalaSkinBridgeListener.js
@@ -76,13 +76,15 @@ OoyalaSkinBridgeListener.prototype.onSeekStarted = function(e) {
   });
 };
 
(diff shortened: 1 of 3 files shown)


brol1dev:
Since you worked on refactoring the SkinSDK in PR #687 I think we need to hold on this one until the other one is merged and change the code syntax of this one to be the same as the new standard you applied when refactoring.

akshaydch11:
We have RN upgrade #687 is merged now. Please update PR to be compatible with that.

brol1dev (changes requested):
@atopilskii can you update the PR to use the new syntax of react native that is being used now in the dev branch?

PR #689, opened 2018-06-15
[React-Native-Update]Removed comments and unused imports

IvanSakharovskii:
added google maven to build repository. Checked that skinSDK library is building, tried to put it locally in sample apps, all works fine.

### DIFF
--- a/sdk/android/build.gradle
+++ b/sdk/android/build.gradle
@@ -13,6 +13,7 @@ buildscript {
 
 allprojects {
     repositories {
+        maven { url "https://maven.google.com" }
         jcenter()
         google()
     }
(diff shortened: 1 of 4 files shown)


sdk/android/build.gradle
```diff
@@ -13,6 +13,7 @@ buildscript {
 
 allprojects {
     repositories {
+        maven { url "https://maven.google.com" }
```
brol1dev, line 16: why do we need it? Isn't `google()` the same as this?

brol1dev (commented):
Just have one question about the maven repo inclusion.

sdk/android/build.gradle
IvanSakharovskii: From google documentation: 

```
 If you're using a version of Gradle lower than 4.1, you must instead use:
 maven {
     url 'https://maven.google.com'
 }
```

https://developer.android.com/studio/build/dependencies

brol1dev (approved):
LGTM

akshaydch11:
LGTM

akshaydch11 (approved):

PR #696, opened 2018-07-10
[PLAYER-3933] 

cacel:
Remove moreOptions button when Ad is shown


cacel:
Solving conflicts first with dev

PR #704, opened 2018-07-19
[PLAYER-3954] Enable in manifest closed captions

orobles-ooyala:


sdk/android/skin/src/main/java/com/ooyala/android/skin/OoyalaSkinPlayerObserver.java
```diff
@@ -218,11 +220,27 @@ private void bridgeAdPodStartedNotification() {
   }
 
   private void bridgeLiveCCChangedNotification(Object data) {
+    Map<String, String> map = (Map<String, String>) data;
```
brol1dev: can `data` be null here too? Better check, I think.

brol1dev (commented):
Added a question. I know is not code that you touched, but the one you created has a null check which I think should be added to the other one too.

sdk/android/skin/src/main/java/com/ooyala/android/skin/OoyalaSkinPlayerObserver.java
orobles-ooyala: Done

brol1dev (approved):
LGTM

akshaydch11 (approved):

PR #705, opened 2018-07-23
[PLAYER-3669] Implement skip forward and backward buttons

IvanSakharovskii:


maxkupetskii (approved):

SZadorozhnyiEpam (approved):

sdk/react/constants.js
```diff
@@ -129,6 +129,8 @@ const VIEW_ACCESSIBILITY_NAMES = {
   CC_CELL: 'Subtitle cell. Tap twice to choose this subtitles',
   PROGRESS_BAR: 'Progress bar. Use two fingers to adjust the progress value',
   VOLUME_BAR: 'Volume bar. Use two fingers to adjust the volume value',
+  FORWARD_BUTTON: 'Forward button. Press twice to seek forward on ',
+  BACKWARD_BUTTON: 'Backward button. Press twice to seek backward on '
```
akshaydch11: missing ',' here

akshaydch11 (changes requested):

sdk/react/ooyalaSkinPanelRenderer.js
```diff
@@ -132,7 +132,8 @@ OoyalaSkinPanelRenderer.prototype.renderVideoView = function() {
         upNext: this.skin.props.upNext,
         icons: this.skin.props.icons,
         adScreen: this.skin.props.adScreen,
-        live: this.skin.props.live
+        live: this.skin.props.live,
+        seekFwd: this.skin.props.seekFwd
```
akshaydch11: Use skipcontrols instead if adding new config in skin.
https://github.com/ooyala/skin-config/blob/master/skin.json#L181

sdk/react/panels/StartScreen.js
```diff
@@ -60,6 +59,14 @@ class StartScreen extends React.Component {
             pause: {
               icon: this.props.config.icons.pause.fontString,
               fontFamily: this.props.config.icons.pause.fontFamilyName
+            },
+            seekForward: {
+              icon: this.props.config.icons.seekForward.fontString,
```
akshaydch11: use replay icon wrapper around seek value use : https://github.com/ooyala/skin-config/blob/master/skin.json#L317

akshaydch11:
We need to use skin config similar to what web is using. I have added more comments. 
[See web UI here](http://debug.ooyala.com/ea/index.html?ec=p4eTdtZTE63eTyP0lTad7S2yA5HN8uhM&pbid=39412bb0d62d41239ed5e4cd048d70a6&pcode=tsbmYyOjRH6dK3rmGxeTP2uleUj_&core_player=%2F%2Fplayer.ooyala.com%2Fstatic%2Fv4%2Fproduction%2Flatest%2Fcore.min.js&video_plugins=%2F%2Fplayer.ooyala.com%2Fstatic%2Fv4%2Fproduction%2Flatest%2Fvideo-plugin%2Fmain_html5.min.js%0A%2F%2Fplayer.ooyala.com%2Fstatic%2Fv4%2Fproduction%2Flatest%2Fvideo-plugin%2Fbit_wrapper.min.js&html5_skin=%2F%2Fplayer.ooyala.com%2Fstatic%2Fv4%2Fproduction%2Flatest%2Fskin-plugin%2Fhtml5-skin.min.js&skin_asset=%2F%2Fplayer.ooyala.com%2Fstatic%2Fv4%2Fproduction%2Flatest%2Fskin-plugin%2Fhtml5-skin.min.css&skin_config=%2F%2Fplayer.ooyala.com%2Fstatic%2Fv4%2Fproduction%2Flatest%2Fskin-plugin%2Fskin.json&ad_plugin=%2F%2Fplayer.ooyala.com%2Fstatic%2Fv4%2Fproduction%2Flatest%2Fad-plugin%2Fgoogle_ima.min.js&options=%7B%22autoplay%22%3Atrue%2C%22initialVolume%22%3A0%2C%22loop%22%3Atrue%2C%22skin%22%3A%7B%22inline%22%3A%7B%22skipControls%22%3A%7B%22enabled%22%3Atrue%7D%7D%7D%7D&trackingLevel=unset)

sdk/react/panels/StartScreen.js
IvanSakharovskii: Related to this task: https://jira.corp.ooyala.com/browse/PLAYER-3650 ,we have new fonts with new buttons. So, i just added them

IvanSakharovskii:
@akshaydch11 @rcvrgs Updated PR, please check.

sdk/react/widgets/VideoViewPlayPause.js
```diff
@@ -133,19 +146,52 @@ class VideoViewPlayPause extends React.Component {
...
+    }
+    return this._renderButton(PLAY);
+  };
+
   _renderButton = (name) => {
     const fontStyle = {fontSize: this.props.fontSize, fontFamily: this.props.icons[name].fontFamily};
     const opacity = {opacity: this.state[name].animationOpacity};
     const animate = {transform: [{scale: this.state[name].animationScale}]};
-    const buttonColor = {color: this.props.buttonColor == null? "white": this.props.buttonColor};
+    const buttonColor = {color: this.props.buttonColor === null ? "white" : this.props.buttonColor};
```
akshaydch11: this is failing and making icon color black

akshaydch11:
Adding update which we talked over chat. So that others are aware of it.
Issues I observed
1. Icons turning black
2. I am not seeing skip fwd and backward buttons on UI
3. Also are we respecting `"enabled": false` from `skipControls` ? if I set it true then app crashes ?

akshaydch11:
Reverified again with latest update. Now it works as expected.

akshaydch11 (approved):

sdk/react/constants.js
```diff
@@ -129,6 +129,8 @@ const VIEW_ACCESSIBILITY_NAMES = {
   CC_CELL: 'Subtitle cell. Tap twice to choose this subtitles',
   PROGRESS_BAR: 'Progress bar. Use two fingers to adjust the progress value',
   VOLUME_BAR: 'Volume bar. Use two fingers to adjust the volume value',
+  FORWARD_BUTTON: 'Forward button. Press twice to seek forward on ',
```
pilievOoyala: Small change: take out the "on" and the extra spaces at the end. For both of these descriptions.

sdk/react/panels/StartScreen.js
```diff
@@ -10,20 +9,20 @@ import {
...
 } from 'react-native';
 
-var Utils = require('../utils');
-var Log = require('../log');
-var styles = Utils.getStyles(require('./style/startScreenStyles.json'));
-var Constants = require('../constants');
-var {
+const Utils = require('../utils');
+const Log = require('../log');
+const styles = Utils.getStyles(require('./style/startScreenStyles.json'));
```
pilievOoyala, line 14: I know it was there before but is there a reason to not start this var name with capital S?  All the other ones are capital. Or should we make all the other lowercase? @aeng7 @dallen42  What do you guys think??

sdk/react/panels/videoView.js
```diff
@@ -125,6 +125,20 @@ class VideoView extends React.Component {
...
+  onSeekPressed = (isForward) => {
+    const seekValue = (isForward) ? this.props.config.skipControls.skipForwardTime :  this.props.config.skipControls.skipBackwardTime * (-1);
+    const currentPlayhead = this.props.playhead;
+    let resultedPlayhead = currentPlayhead + seekValue;
+    if (resultedPlayhead < 0) {
+      resultedPlayhead = 0;
+    } else if (resultedPlayhead > this.props.duration) {
+      resultedPlayhead = this.props.duration;
+    }
+    const resultedPlayheadPercent = resultedPlayhead / this.props.duration;
```
pilievOoyala: Is the duration always > 0? Is there a way to call this in between videos? I'm worried about an edge case of dividing by 0.

sdk/react/widgets/VideoViewPlayPause.js
```diff
@@ -102,22 +115,22 @@ class VideoViewPlayPause extends React.Component {
...
     this.state.widget.animationOpacity.setValue(this.props.showButton ? 1 : 0);
-    this.setState({inAnimation:false});
+    this.setState({inAnimation: false});
     this.syncButtons(this.props.playing);
   };
 
   // Animations for play/pause transition
   animatePlayButton = () => {
     this.setState({inAnimation: true});
-    this.state.play.animationScale.setValue(1);
```
pilievOoyala, line 112: I'm slightly confused about these changes. You are reversing the animation to go from invisible to visible instead of the other way around. But why?

sdk/react/panels/StartScreen.js
IvanSakharovskii: @pilievOoyala because style in this case is object, and other variables is Components with functions

sdk/react/constants.js
IvanSakharovskii: @pilievOoyala This space is required to make a composite string. In accessibility it will be something like: `Forward button. Press twice to seek forward on 10 seconds`

sdk/react/panels/videoView.js
IvanSakharovskii: @pilievOoyala You a right. I'm adding check

sdk/react/widgets/VideoViewPlayPause.js
IvanSakharovskii: @pilievOoyala before animation was wrong. It started from size x1 and animated to size x2. After animation's end function `showPlayButton` was called, where we changed scale from value `x2` to value `x1`. And on this behavior we had a non-animated play button size changes. It looked a bit strange.

sdk/react/panels/videoView.js
IvanSakharovskii: @pilievOoyala added check duration on 0 value.

sdk/react/widgets/FwdButton.js
```diff
@@ -0,0 +1,70 @@
...
+  Animated,
+  View,
+  Text
+} from 'react-native';
+
+const AccessibilityUtils = require('../accessibilityUtils');
+const Constants = require('../constants');
+const styles = require('../utils').getStyles(require('./style/RectButtonStyles.json'));
+
+class FwdButton extends React.Component {
```
akshaydch11: instead of FwdButton can we change it to SkipButton. SkipButton seems more relevant with the feature.

IvanSakharovskii: @akshaydch11 Renamed

sdk/react/constants.js
brol1dev: @IvanSakharovskii, we could add the space like this: `'Press twice to seek forward on' + '  ' + x + 'seconds'`. 

I assume that right now it is `'Press twice to seek forward on ' + x + 'seconds'`.

Understand why you want the space, not sure which is the right approach, but yea would suggest to keep the Constants without spaces.

At least for now remove the "on" :)

sdk/react/widgets/VideoViewPlayPause.js
```diff
@@ -133,19 +146,52 @@ class VideoViewPlayPause extends React.Component {
...
+    }
+    return this._renderButton(PLAY);
+  };
+
   _renderButton = (name) => {
     const fontStyle = {fontSize: this.props.fontSize, fontFamily: this.props.icons[name].fontFamily};
     const opacity = {opacity: this.state[name].animationOpacity};
     const animate = {transform: [{scale: this.state[name].animationScale}]};
-    const buttonColor = {color: this.props.buttonColor == null? "white": this.props.buttonColor};
+    const buttonColor = {color: this.props.buttonColor == null ? "white" : this.props.buttonColor};
```
brol1dev: should it be `this.props.buttonColor === null `?

sdk/react/widgets/VideoViewPlayPause.js
```diff
@@ -133,19 +146,52 @@ class VideoViewPlayPause extends React.Component {
...
+        accessible={false}
+        style={[styles.buttonTextStyle, fontStyle, buttonColor, this.props.buttonStyle, animate, opacity]}>
+        {this.props.icons[name].icon}
+      </Animated.Text>
+    );
+  };
+
+  _renderSeekButton = (name, iconScale) => {
+    if (!this.props.showSeekButtons) {
+      return <View/>
```
brol1dev, line 173: What happens if you return nothing or null?

sdk/react/widgets/VideoViewPlayPause.js
akshaydch11: @rcvrgs For some reason === does not work here. It fails and changes color to black.

sdk/react/widgets/VideoViewPlayPause.js
IvanSakharovskii: It's ok to return nothing or null here, I just wanted to prevent  mistakes in the future, because for some android versions we could get error if we have null in the view hierarchy.

sdk/react/widgets/VideoViewPlayPause.js
IvanSakharovskii: @rcvrgs in current case for props we have `undefined` and `===` returns `false` on this check. I've changed it to `!!this.props.buttonColor`, it should work correctly.

sdk/react/constants.js
IvanSakharovskii: @rcvrgs Ok, done :)

akshaydch11:
@IvanSakharovskii  Found one more issue. 
As per current settings, if accessibility is ON we are not showing playpause screen. Hence it is also not showing skip buttons.

brol1dev (approved):
I'm approving and merging to let QA start testing. As @akshaydch11 mentions, we need to implement it well for Accessibility or let us know if we're missing details that we're not aware of.

PR #713, opened 2018-08-02
[PLAYER-3660, PLAYER-4061, PLAYER-4055, PLAYER-4054] Added accessibility for forward/backward buttons and changed PLAY/PAUSE behavior

IvanSakharovskii:
I removed all animation logic from PLAY/PAUSE button, because it didn't work, and all animated was added in button params.

### DIFF
--- a/sdk/react/constants.js
+++ b/sdk/react/constants.js
@@ -129,11 +129,11 @@ const VIEW_ACCESSIBILITY_NAMES = {
   CC_CELL: 'Subtitle cell. Tap twice to choose this subtitles',
   PROGRESS_BAR: 'Progress bar. Use two fingers to adjust the progress value',
   VOLUME_BAR: 'Volume bar. Use two fingers to adjust the volume value',
-  FORWARD_BUTTON: 'Forward button. Press twice to seek forward',
-  BACKWARD_BUTTON: 'Backward button. Press twice to seek backward',
+  FORWARD_BUTTON: 'Forward button. Tap twice to seek forward',
+  BACKWARD_BUTTON: 'Backward button. Tap twice to seek backward',
   ENTER_FULLSCREEN: 'Enter Fullscreen mode button selected. Double tap to activate.',
   EXIT_FULLSCREEN: 'Exit Fullscreen mode button selected. Double tap to activate.',
-  PLAY_PAUSE_BUTTON: "button. Press twice to",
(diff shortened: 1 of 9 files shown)


sdk/react/constants.js
```diff
@@ -133,6 +133,7 @@ const VIEW_ACCESSIBILITY_NAMES = {
   BACKWARD_BUTTON: 'Backward button. Press twice to seek backward',
   ENTER_FULLSCREEN: 'Enter Fullscreen mode button selected. Double tap to activate.',
   EXIT_FULLSCREEN: 'Exit Fullscreen mode button selected. Double tap to activate.',
+  PLAY_PAUSE_BUTTON: "button. Press twice to",
```
brol1dev: Use Tap instead of Press, do the same for the `FORWARD_BUTTON` and `BACKWARD_BUTTON`

sdk/react/index.ios.js
```diff
@@ -61,7 +61,7 @@ class OoyalaSkin extends React.Component {
   };
 
   componentWillMount() {
-    console.log("componentWillMount");
+    Log.verbose("componentWillMount");
```
brol1dev: I think we can remove this log, not sure how useful it is for us. What do you think?

brol1dev (changes requested):
Have some a couple of comments

brol1dev:
@IvanSakharovskii do you know the Jira tickets related to this PR?

IvanSakharovskii:
@rcvrgs Yes, this PR related to PLAYER-4061, PLAYER-4055, PLAYER-4054 tasks. Also user story for it is PLAYER-3660. I will fix comments and update PR. If you find something else please add

akshaydch11:
@IvanSakharovskii  if possible can you take a look at PLAYER-4052 and PLAYER-4052 something in seek pressed calculation going wrong when video length is very long.

brol1dev (approved):
Tested on an iOS device. Works fine in my opinion.

IvanSakharovskii:
@akshaydch11 I'll add comment to task, that related to seek. Current behavior is wrong because of our playhead is converting to percent and in native code converting back to the playhead and from percent we check only 2 decimal places, so i added a round for it. When we have a long video our round value converting to the playhead with a big value. I will add example in task comment a bit later

akshaydch11 (approved):
LGTM

PR #714, opened 2018-08-03
[PLAYER-4076, PLAYER-4077] Fixed behavior if skip time value was out of range

IvanSakharovskii:


sdk/react/panels/videoView.js
```diff
@@ -121,7 +121,13 @@ class VideoView extends React.Component {
   };
 
   onSeekPressed = (isForward) => {
-    const seekValue = (isForward) ? this.props.config.skipControls.skipForwardTime : this.props.config.skipControls.skipBackwardTime * (-1);
+    let configSeekValue = (isForward) ? this.props.config.skipControls.skipForwardTime : this.props.config.skipControls.skipBackwardTime;
+    if (configSeekValue > 99) {
```
SZadorozhnyiEpam: Would be great to extract `1` and `99` since we are using it in 2 different files

sdk/react/panels/videoView.js
```diff
@@ -121,7 +121,13 @@ class VideoView extends React.Component {
...
+      configSeekValue = 99;
+    } else if (configSeekValue < 1) {
+      configSeekValue = 1;
+    }
```
AlinaVoronkovaEpam: Can we use smth like this:
Number.prototype.clamp = function(min, max) {
  return Math.min(Math.max(this, min), max);
};
instead of if-else statement?

AlinaVoronkovaEpam (changes requested):

sdk/react/panels/videoView.js
IvanSakharovskii: @AlinaVoronkovaEpam updated, check please

AlinaVoronkovaEpam (approved):

MarkEPAM (approved):

akshaydch11 (approved):

sdk/react/utils.js
```diff
@@ -230,13 +231,8 @@ var Utils = {
...
 
-  makeAccessibilityLabelWithParams: function(viewName, baseAccessibleLabel, type) {
-    switch (type) {
-      case "seek_views":
-        return `${viewName}. ${baseAccessibleLabel}`;
-      default:
-        return value;
-    }
+  restrictSeekValueIfNeeded: function (seekValue) {
+    return Math.min(Math.max(Constants.VALUES.MIN_SKIP_VALUE, seekValue), Constants.VALUES.MAX_SKIP_VALUE);
```
pilievOoyala: Per the style guide can you stick this in a temp variable and return that.

pilievOoyala:
LGTM aside from the one style issue. You can merge when you make that change.

brol1dev (approved):
LGTM

PR #717, opened 2018-08-06
[Player-4078] Removed redundant component name

SZadorozhnyiEpam:
https://jira.corp.ooyala.com/browse/PLAYER-4078
TalkBack announces component type after accessibility label. 
Accessibility label gives enough information about buttons so component type, in this case, is redundant.


akshaydch11 (approved):

pilievOoyala:
LGTM

PR #721, opened 2018-08-07
[Player-4060] Seek buttons visibility in live video

SZadorozhnyiEpam:
Performed changes:
For DVR:
If scrubber in a live region then fast forward button will have grey color and not react to any touches.
For Live without DVR support:
Not yet implemented since we don't have a clear indicator of how to detect if Live has DVR or not.
But since we have `forceDvrDisabled` in skin.config and if this value is set to true then seek buttons will be invisible for the user.


sdk/react/panels/videoView.js
```diff
@@ -272,6 +272,8 @@ class VideoView extends React.Component {
 
   _renderPlayPause = (show) => {
     const iconFontSize = ResponsiveDesignManager.makeResponsiveMultiplier(this.props.width, UI_SIZES.VIDEOVIEW_PLAYPAUSE);
+    const seekVisible = !this.props.config.live.forceDvrDisabled;
+    const notInLiveRegion = this.props.playhead <= this.props.duration * 0.95;
```
AlinaVoronkovaEpam: what does 0.95 mean? Could you move it to constants?

AlinaVoronkovaEpam (commented):
LGTM. Please make minor changes. See my comment.

sdk/react/panels/videoView.js
SZadorozhnyiEpam: 0.95 is a value from which player thinks that Live video should display Live icon, so we also disable FF button to synchronize them both.

>Could you move it to constants? 
Will do

akshaydch11 (approved):
LGTM

sdk/react/panels/videoView.js
brol1dev: agree with @AlinaVoronkovaEpam , wherever you found the usage of 0.95 please move it all to a constant so that it's easier to understand.

sdk/react/widgets/VideoViewPlayPause.js
```diff
@@ -144,8 +147,8 @@ class VideoViewPlayPause extends React.Component {
   render() {
     const seekButtonScale = 0.5;
     const playPauseButton = this._renderPlayPauseButton();
-    const backwardButton = this._renderSeekButton(BACKWARD, seekButtonScale);
-    const forwardButton = this._renderSeekButton(FORWARD, seekButtonScale);
+    const backwardButton = this._renderSeekButton(BACKWARD, seekButtonScale, true);
```
brol1dev, line 156: When forceDvrDisabled is on, both seek buttons should be hidden because you shouldn't be able to scrub back in time. Add the logic for this button too.

brol1dev (changes requested):
Overall it looks good, but I added a couple of comments that I believe are important.

sdk/react/widgets/VideoViewPlayPause.js
brol1dev: my bad, realized the logic is correct here. Nothing to do. Sorry @SZadorozhnyiEpam

brol1dev (approved):
LGTM. Just added the comment about the `0.95` constant. I'll try to add the change on my side so we can create a new RC.

Thanks

sdk/react/widgets/VideoViewPlayPause.js
```diff
@@ -112,15 +114,15 @@ class VideoViewPlayPause extends React.Component {
...
+  _renderSeekButton = (name, iconScale, active) => {
+    if (!this.props.showSeekButtons || !this.props.seekEnabled) {
       return <View/>
     }
     const fontStyle = {fontSize: this.props.fontSize * iconScale, fontFamily: this.props.icons[name].fontFamily};
     const sizeStyle = {width: this.props.buttonWidth, height: this.props.buttonHeight};
     const opacity = {opacity: this.state.skipButtons.animationOpacity};
     const animate = {transform: [{scale: this.state.skipButtons.animationScale}]};
-    const buttonColor = {color: !!this.props.buttonColor ? this.props.buttonColor : "white"};
+    const buttonColor = {color: active ? !!this.props.buttonColor ? this.props.buttonColor : "white" : "gray"};
```
pilievOoyala: Please don't nest turnary statements. They are hard to follow. Can you separate out this logic into if statements and then assign buttonColor the result.

brol1dev: changed it

brol1dev:
@pilievOoyala, @akshaydch11 please review again. Added a couple of commits

akshaydch11:
Looks good now

pilievOoyala:
LGTM

PR #723, opened 2018-08-08
[PLAYER-4060] Seek visibility for VOD videos fix

SZadorozhnyiEpam:
4060 has been reopened - fix affected not only live videos but also VODs. With `forceDvrDisabled = true` seek buttons were hidden.
Basically, I forgot to add a condition -  when DVR Forced hide seek buttons **only** for live videos.


AlinaVoronkovaEpam (approved):

akshaydch11 (approved):

brol1dev (approved):
LGTM

PR #725, opened 2018-08-09
[PLAYER-4131] handled null pointer exception

akshaydch11:


sdk/android/skin/src/main/java/com/ooyala/android/skin/OoyalaSkinLayoutController.java
```diff
@@ -353,7 +353,7 @@ void updateBridgeWithCurrentState() {
 
     if (_package.getBridge() == null) {
       DebugMode.logE(TAG, "Got onMounted from React, but bridge is not yet available? Invalid State");
-    } else {
+    } else if (null != queuedEvents) {
```
brol1dev: per code style guide, we shouldn't be defining this kind of null checks. Do variable first then null. Personally, I prefer it like it is right now but we would have to change the style guide to use it this way.

brol1dev (changes requested):

brol1dev (approved):
LGTM

PR #727, opened 2018-08-10
Player 4093

maxkupetskii:
-skip buttons now have delay and sum up all presses in order to maintain smooth experience


sdk/react/panels/videoView.js
```diff
@@ -121,10 +121,13 @@ class VideoView extends React.Component {
...
   };
 
-  onSeekPressed = (isForward) => {
-    let configSeekValue = (isForward) ? this.props.config.skipControls.skipForwardTime : this.props.config.skipControls.skipBackwardTime;
-    configSeekValue = Utils.restrictSeekValueIfNeeded(configSeekValue);
-    const seekValue = (isForward) ? configSeekValue : configSeekValue * (-1);
+  onSeekPressed = (skipCountValue) => {
+    if (skipCountValue == 0) { return null; }
+
+    let configSeekValue = (skipCountValue > 0) ? this.props.config.skipControls.skipForwardTime : this.props.config.skipControls.skipBackwardTime;
```
MarkEPAM, line 127: Can we use ```const``` instead ```let```?

maxkupetskii: I didn't change this part, so it was already `let`.

maxkupetskii: If we keep the `restrictSeekValueIfNeeded` disabled then we can use `const`.

MarkEPAM (approved):

sdk/react/panels/videoView.js
```diff
@@ -121,10 +121,13 @@ class VideoView extends React.Component {
...
+    //configSeekValue = Utils.restrictSeekValueIfNeeded(configSeekValue);
```
brol1dev, line 128: why do we remove this? We need it so that we accept values between 1-99 only.

brol1dev (changes requested):

sdk/react/panels/videoView.js
maxkupetskii: It's a temporal solution because in the ticket we discussed another behavior.

Without this PR we can skip only in interval of 1..99, but if we are to sum up all taps on skip button it's possible that we will get a value larger than 99. 

We have a check further that it's still in asset duration.

pilievOoyala:
LGTM aside from the other questions raised.

brol1dev (approved):
Approving for now but we still need to value validation.

PR #728, opened 2018-08-13
[PLAYER-4144] Fixed appearing of a black screen on drag/hold motion.

SZadorozhnyiEpam:
https://jira.corp.ooyala.com/browse/PLAYER-4144

The bug appeared after https://jira.corp.ooyala.com/browse/PLAYER-4062 fix.

After my changes, both issues seem to be resolved on Android os. I assume since it is react-native, ios should be fixed too.

### DIFF
--- a/sdk/react/widgets/VideoViewPlayPause.js
+++ b/sdk/react/widgets/VideoViewPlayPause.js
@@ -89,10 +89,6 @@ class VideoViewPlayPause extends React.Component {
     }
   };
 
-  onPressBackground = () => {
-    this.props.onPress(BUTTON_NAMES.RESET_AUTOHIDE);
-  }
-
   onSkipPress = (isForward) => {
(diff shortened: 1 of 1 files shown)


MarkEPAM (approved):

akshaydch11 (approved):

brol1dev (commented):
I believe that if we merge this PR it will break https://jira.corp.ooyala.com:8443/browse/PLAYER-4062. We need to confirm and if that's the case then we'll have to find another fix. Why is the reason that the code that you remove causes a black screen to show?

brol1dev (commented):
My bad, I just saw your comments that you saw that PLAYER-4062 is not broken. Will verify in both platforms and see how it works. After that I'll merge.

brol1dev (approved):
LGTM

PR #730, opened 2018-08-13
[PLAYER-4142] Fixed crash in case of no closedCaption

akshaydch11:


sdk/react/panels/ItemSelectionList.js
```diff
@@ -21,11 +21,12 @@ class ItemSelectionList extends React.Component {
...
-        this.props.itemRender(item, index)
-    ));
-
-    return (renderedItems);
+    if (typeof(items) !== "undefined") {
+      const renderedItems = items.map((item, index) => (
+          this.props.itemRender(item, index)
+      ));
+      return (renderedItems);
+    }
```
brol1dev: doesn't this one need another return value in case it's `undefined`?

brol1dev (commented):
Added one question

sdk/react/panels/ItemSelectionList.js
pilievOoyala: I think it will return null if you don't specify anything. But, we want to be explicit, so please add a return statement for the undefined case.

akshaydch11: @rcvrgs Yeah thought about it but not sure what we can return for undefined value. Currently, it shows blank screen. 
Output of renderItems() is passed to ScrollView which is expected not show anything.

Tried to "return items" and gave me same result.

akshaydch11: @rcvrgs @pilievOoyala  Updated now.

brol1dev (approved):
LGTM

ipadilla4 (approved):
LGTM

PR #731, opened 2018-08-13
[PLAYER-4076][PLAYER-4077] Re-add filtering seek values between 1-99

brol1dev:
I didn't see any issue in re-adding this filter option. I think SPB was confused about the requirement. It's ok to go beyond 99 if you sum up the seek actions, but a single seek action should not be out of the 1-99 range.


akshaydch11 (approved):
LGTM

pilievOoyala:
LGTM

PR #733, opened 2018-08-14
[PLAYER-4154][PLAYER-4062] Calculate buttons container position dynamically

brol1dev:
Depending on the buttons that will be shown, we calculate the position of the buttons that will be shown in the middle of the video area. This logic was similar before the skip buttons were introduced, just modified it to work with skip buttons too.


akshaydch11 (approved):
Have you tested on android device ? 

LGTM otherwise

brol1dev:
Tested with an iPad and an Android device. But only used one per platform. Hopefully there are no special case devices, @akshaydch11

ipadilla4 (approved):
LGTM

PR #734, opened 2018-08-15
[PLAYER-4053] Send raw seek percent value

brol1dev:
This fix is only working on iOS atm. We need to change some Android SDK methods to make it work on Android too.


pilievOoyala:
LGTM. but is there any place where we display the number that would still need to be rounded? We may have to round it before displaying it in the ui.

brol1dev:
I don't think at this place @pilievOoyala . The flow of this is
1. Calculate the percentage value from seek bar and send it to the core iOS SDK
2. Core sdk converts it from percentage to a real playhead value and send the message to the `OoyalaPlayer`
3. Player updates itself and send message bus events.
4. The skin SDK listens for a new playhead event and updates the UI accordingly.

brol1dev:
I'm self merging this one we need it now to unblock the release.

PR #735, opened 2018-08-17
[PBA-3979 REFACTOR] Global refactoring of OoyalaSkinSDK

maxkupetskii:
# OoyalaSkinSDK global refactoring
I was working on a ticket _PBA-3979_ and realized that some deprecated RN method could cause an issue also related to that ticket. Eventually, I've discovered that some code could cause a fatal error for a client and/or confusion for any developer involved. Eventually this lead to global refactoring.

#### Briefly
Now SDK has a model, native UI part is responsible for UI only, RN part is rewritten in a modern way, JS part got rid of ambiguity, modern Objective-C applied, code style applied.

### Added
- RN part completely refactored. It includes:
  - `OOReactSkinModel`. It's a model in architectural sense. It's responsible for sending events to RN, for creating RCTRootView instance and for handling methods from RN
  - `OOReactSkinEventsEmitter`. It's a RN module responsible for emitting events only. Now in JS it's created as a `NativeEventEmitter`. It's also populates all supported events (sending not populated event will cause RN to crash) and handles `OOQueueEvent` array.
  - `OOReactSkinBridge` introduced as a subclass of `RCTBridge`. Since all RN modules are being created by RN automatically this bridge can return an instance of `OOReactSkinEventsEmitter` module.
  - `OOReactSkinBridgeModule` protocol introduced. Any RN module should implement it.
  - `OOReactSkinBridgeModuleMain` module introduced. It exports available methods to RN and handles them with `OOReactSkinModelDelegate` delegate.

- `OOSkinViewControllerDelegate` protocol added in order to handle UI interactions from the model

### Changed
- `OOSkinViewController` now handles only UI logic. It's not passed anywhere and all business logic is implemented inside `OOReactSkinModel`
- `OOSkinPlayerObserver` uses `OOReactSkinModel` to send events

- Modern Objective-C applied
  - All headers were cleaned from unnecessary imports
  - All headers now use forward class declarations
  - `#define` moved to `static NSString *const`
  - All string literals were moved to constants
  - All files were refactored to follow one style guide
  - `CGRectGetWidth` and `CGRectGetHeight` used where possible
  - `NSNumber` are being inited with modern `@(var)` syntax
  - `NSDictionary` are being inited with modern `@{...}` syntax

### Removed
- `OOReactiBridge` removed, all functionality split
- `OOQueeudEvent` now in `React` section
- `OOSkinViewController+Internal.h` removed
- `disableBuiltInAdLearnMoreButton` method from `OOSkinViewController` because it's present in `OOSkinPlayerObserver`


brol1dev:
WOW, this is a big change, in general I like it. @maxkupetskii let me spend some time to really check it out. My initial thoughts are do we also need to change logic on Android? I would assume yes.

maxkupetskii:
@rcvrgs that's fine.
I can't say anything about Android but this PR doesn't affect Android SDK. If we want to have similar architecture on both platforms then probably some changes should be made.

sdk/iOS/OoyalaSkinSDK/OOUpNextManager.m
```diff
@@ -48,10 +64,10 @@ - (void)setNextVideo:(NSDictionary *)nextVideo {
 
     // After the a new video has been set, let react know that isDismissed
     // is now false.
-    [self.ooBridge sendDeviceEventWithName:@"upNextDismissed" body:@{@"upNextDismissed" : [NSNumber numberWithBool:[self isDismissed]]}];
+    [self.ooReactSkinModel sendEventWithName:upNexDismissedKey body: @{upNexDismissedKey: @(self.isDismissed)}];
```
pilievOoyala: Is this misspelled? upNexDismissedKey should be upNextDismissedKey

pilievOoyala: Can you please change it regardless, for consistency with the rest of the code.

sdk/iOS/OoyalaSkinSDK/React/OOReactSkinBridgeModuleMain.m
```diff
@@ -0,0 +1,172 @@
...
+static NSString *socialShareButtonName  = @"SocialShare";
+static NSString *fullscreenButtonName   = @"Fullscreen";
+static NSString *learnMoreButtonName    = @"LearnMore";
+static NSString *skipButtonName         = @"Skip";
+static NSString *moreOptionButtonName   = @"More";
+static NSString *upNextDismiss          = @"upNextDismiss";
+static NSString *upNextClick            = @"upNextClick";
+static NSString *adIconButtonName       = @"Icon";
+static NSString *adOverlayButtonName    = @"Overlay";
+static NSString *stereoscopicButtonName = @"stereoscopic";
```
pilievOoyala, line 54: Why do some of these values start with a capital letter and some don't?  Is this just inconsistency with how they were declared elsewhere in the code?

sdk/iOS/OoyalaSkinSDK/React/OOReactSkinEventsEmitter.m
```diff
@@ -0,0 +1,105 @@
+/**
+ * @file       OOReactSkinEventsEmitter.m
+ * @class      OOReactSkinEventsEmitter OOReactSkinEventsEmitter.m "OOReactSkinEventsEmitter.m"
+ * @brief      OOReactSkinEventsEmitter
+ * @details    OOReactSkinEventsEmitter.m in OoyalaSDK
+ * @date       13/8/18
```
pilievOoyala: This is just nitpicking but can you write it as 8/13/18. Then it'll be consistent with the rest of the code.

sdk/iOS/OoyalaSkinSDK/React/OOReactSkinModel.m
```diff
@@ -0,0 +1,352 @@
...
+
+- (void)playPauseFromAdTappedNotification {
+  if (![self.skinControllerDelegate isReactViewInteractionEnabled]) {
+    [self handlePlayPause];
+  }
+}
+
+- (void)handleReplay {
+  [self.player seek:0];
+  [self.player play];
```
pilievOoyala, line 299: Do we need to do anything else here so that ads replay when you watch the video again?

sdk/iOS/OoyalaSkinSDK/ViewControllers/OOSkinViewController.m
```diff
@@ -3,68 +3,44 @@
 //  OoyalaSkin
 //
 //  Created by Zhihui Chen on 4/16/15.
-//  Copyright (c) 2015 Facebook. All rights reserved.
```
pilievOoyala, line 6: Excellent catch

sdk/iOS/OoyalaSkinSDK/ViewControllers/OOSkinViewController.m
```diff
@@ -573,87 +413,20 @@ - (CGRect)videoViewFrame {
 - (void)toggleStereoMode {
   dispatch_async(dispatch_get_main_queue(), ^{
     _isVRStereoMode = !_isVRStereoMode;
-    
-    if (_isVRStereoMode) {
-      [self enterStereoMode];
-    } else {
-      [self exitStereoMode];
-    }
+    self.isVRStereoMode ? [self enterStereoMode] : [self exitStereoMode];
```
pilievOoyala: Please keep the old style of if statement here. It's slightly easier to debug and set a breakpoint for debugging.  Same for the line below.

pilievOoyala:
Overall looks great. It's much easier to follow. It's a lot of code so I can't dig into it as much without spending hours on it.  I just found naming issues and a couple of places I know we've had problems before on the web side (ie, replay).  @rcvrgs @akshaydch11 Please review again and if we are going to merge it  let's do it sooner than later so we can test thoroughly.

sdk/iOS/OoyalaSkinSDK/OOUpNextManager.m
maxkupetskii: Done

sdk/iOS/OoyalaSkinSDK/React/OOReactSkinBridgeModuleMain.m
maxkupetskii: It's all in pure RN `.js` files. It's possible to do, but it will affect Android SDK as well.
I guess it's better to do as a separate task.

sdk/iOS/OoyalaSkinSDK/React/OOReactSkinModel.m
maxkupetskii: @cacel could you please answer?

sdk/iOS/OoyalaSkinSDK/ViewControllers/OOSkinViewController.m
maxkupetskii: :)

sdk/iOS/OoyalaSkinSDK/ViewControllers/OOSkinViewController.m
maxkupetskii: Done

maxkupetskii: Okay

grigory9 (approved):

PR #736, opened 2018-08-17
[PBA-6154] Added new logic for auto full screen mode with device rotation changes

MarkEPAM:
- Added new boolean property (autoFullscreenWithRotatedEnabled) for support auto full screen mode when device orientation was changed.
- was deprecated property for manual change full screen mode (in a future release need to set this property as readonly)
- fixed bug with inline animtaion
- fixed bug in task: https://jira.corp.ooyala.com/browse/PLAYER-3979 (need confirmation)

Unit tests not included.
Ticket: https://jira.corp.ooyala.com/browse/PBA-6154

### DIFF
--- a/sdk/iOS/OoyalaSkinSDK/ViewControllers/FullscreenStateController/FullscreenStateController.m
+++ b/sdk/iOS/OoyalaSkinSDK/ViewControllers/FullscreenStateController/FullscreenStateController.m
@@ -11,7 +11,9 @@
 #import "PresentedViewControllerHelper.h"
 
 
-#define FULLSCREEN_ANIMATION_DURATION 0.5
+#pragma mark - Constants
+
+#define FULLSCREEN_ANIMATION_DURATION 0.4
 
 
 @interface FullscreenStateController()
@@ -19,6 +21,7 @@
 @property (nonatomic) UIView *parentView;
 @property (nonatomic) UIView *containerView;
 @property (nonatomic) UIView *videoView;
+@property (nonatomic) CGRect originalVideoViewFrame;
 @property (nonatomic) UIViewController *rootViewController; // Root VC from visible UIWindow
 @property (nonatomic) UIViewController *fullscreenViewController;
 @property (nonatomic) NSOperationQueue *operationQueue; // Queue for fullscreen animation
@@ -52,18 +55,19 @@
 
 #pragma mark - Public functions
 
-- (void)setFullscreen:(BOOL)fullscreen completion:(nullable void (^)())completion {
+- (void)setFullscreen:(BOOL)fullscreen withOrientaionChanges:(BOOL)isOrientaionChanges completion:(nullable void (^)())completion {
   for (NSOperation *operation in self.operationQueue.operations) {
     if (!operation.isExecuting) {
       [operation cancel];
     }
   }
   
-  NSOperation *operation = [[FullscreenStateOperation alloc] initWithFullscreen:fullscreen enterFullscreenBlock:^(void (^animationCompletion)(void)) {
+  NSOperation *operation = [[FullscreenStateOperation alloc] initWithFullscreen:fullscreen
+                                                           enterFullscreenBlock:^(void (^animationCompletion)(void)) {
     if (self.isFullscreen == fullscreen) {
       animationCompletion();
     } else {
-      [self openFullscreenMode:^{
+      [self openFullscreenModeWithOrientaionChanges:isOrientaionChanges completion:^{
         animationCompletion();
       }];
     }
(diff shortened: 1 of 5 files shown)


sdk/iOS/OoyalaSkinSDK/ViewControllers/FullscreenStateController/FullscreenStateController.m
```diff
@@ -11,14 +11,17 @@
 #import "PresentedViewControllerHelper.h"
 
 
-#define FULLSCREEN_ANIMATION_DURATION 0.5
+#pragma mark - Constants
+
+#define FULLSCREEN_ANIMATION_DURATION 0.4
```
maxkupetskii: I think we better use static.

> Constants should be declared as static constants and not #defines unless explicitly being used as a macro

maxkupetskii (changes requested):

sdk/iOS/OoyalaSkinSDK/ViewControllers/FullscreenStateController/FullscreenStateController.m
MarkEPAM: It's old code, I just added pragma mark. We'll change code style after it will be finished.

akshaydch11 (approved):

sdk/iOS/OoyalaSkinSDK/ViewControllers/FullscreenStateController/FullscreenStateController.m
```diff
@@ -31,6 +29,10 @@ @interface FullscreenStateController()
 
 @implementation FullscreenStateController
 
+#pragma mark - Constants
+
+static double const FULLSCREEN_ANIMATION_DURATION = 0.4;
```
brol1dev, line 34: is it 400 millis? why 400?

sdk/iOS/OoyalaSkinSDK/ViewControllers/OOSkinViewController.h
```diff
@@ -28,7 +28,13 @@ extern NSString *const OOSkinViewControllerFullscreenChangedNotification; /* Fir
 /**
  Programatically change the fullscreen mode of the player.
  */
-@property(nonatomic, getter=isFullscreen) BOOL fullscreen;
+@property (nonatomic, getter=isFullscreen) BOOL fullscreen DEPRECATED_MSG_ATTRIBUTE("Setter is deprecated and will be removed in a future version. Use autoFullscreenWithRotatedEnabled instead for enable/disable auto full screen mode after device rotation changes.");
```
brol1dev, line 31: why deprecate the setter? For uses cases in which you just want to go fullscreen by another action which is not rotation it should still be valid, isn't it? It's ok if we add another property (`autoFullscreenWithRotatedEnabled`) just to deal with the fullscreen when rotated.

brol1dev (commented):
Have a couple of questions. Please answer @MarkEPAM

sdk/iOS/OoyalaSkinSDK/ViewControllers/FullscreenStateController/FullscreenStateController.m
MarkEPAM: yes, this value in seconds. Its time period selected, because it's great time combination with iOS rotated animation. We experimented with other values, but this is the best value.

sdk/iOS/OoyalaSkinSDK/ViewControllers/OOSkinViewController.h
MarkEPAM: The default full screen logic may be working not correct with iOS rotation animation, because we can't determine what the full screen animation was initialized with. (caused by rotation or just setter call).

So we can save this implementation, but we can't guarantee a correct animation when customers set the full screen when the device did change orientation.

Also we can add an additional method like this:
```- (void)setFullscreen:(BOOL)fullscreen isCausedByRotation:(BOOL)isCausedByRotation```

brol1dev (approved):
can you review once more, @maxkupetskii? If you think is ok you can merge it.

Thanks.

sdk/iOS/OoyalaSkinSDK/ViewControllers/OOSkinViewController.h
brol1dev: remove the `DEPRECATED_MSG_ATTRIBUTE` for now, we won't remove this setter but will encourage customers to use the new one to deal with fullscreen when rotating.

akshaydch11:
@maxkupetskii if you need any changes please ask @MarkEPAM  to create new PR
merging this to create RC build.

PR #740, opened 2018-08-28
[PLAYER-3700] Added UI for playback speed feature

MarkEPAM:
- added playback speed panel
- added playback speed button 
- added bridge methods for handling playback speed events
Unit tests not included.
Task: https://jira.corp.ooyala.com/browse/PLAYER-3700


sdk/react/controlBar.js
```diff
@@ -33,9 +33,12 @@ var ControlBarWidget = require('./widgets/controlBarWidgets');
 var CollapsingBarUtils = require('./collapsingBarUtils');
 var VolumeView = require('./widgets/VolumeView');
 var ResponsiveDesignManager = require('./responsiveDesignManager');
-
 var styles = Utils.getStyles(require('./style/controlBarStyles.json'));
 
+const constants = {
+  playbackSpeedRatePostfix: "x"
```
pilievOoyala, line 39: This is probably fine but do some languages use a different symbol for multiplier? Should this be overridable?

pilievOoyala:
LGTM.

sdk/react/controlBar.js
pilievOoyala: Disregard this. I asked JP and we don't allow you to customize it on web so you don't need to either.

brol1dev (approved):
LGTM

PR #745, opened 2018-09-03
[PLAYER-3994] Fixed issue when replay button is visible after scrubbing

SZadorozhnyiEpam:
-Added playPause action after onScrub action for EndScreen.
-minor code changes (typo, codestyle)
- `replaybuttonSE` `replaybuttonSW` - unused, deleted

PR #747, opened 2018-09-04
[PLAYER-4134] Replay icon

cacel:
[iOS][Android][OoyalaSkinSampleApp] Replay icon is displayed on seeking back after end of video and play instead replay from sec 0.


brol1dev (changes requested):
There's a conflict in the PR, please fix.

sdk/android/skin/src/main/java/com/ooyala/android/skin/OoyalaSkinBridgeEventHandlerImpl.java
```diff
@@ -104,6 +107,11 @@ public void handleRewind() {
     _player.seek(playheadTime);
   }
 
+  public void handleReplay() {
+    _player.seek(0);
```
pilievOoyala, line 111: Is this enough?  What happens to ads? Do they replay using this method?

brol1dev:
@cacel also there's a related issue that we just merged a fix (#745) Can you check that one out too and see how different it is? I'm not sure if there's a difference or if the tickets are actual duplicates.

sdk/android/skin/src/main/java/com/ooyala/android/skin/OoyalaSkinBridgeEventHandlerImpl.java
cacel: The seek is set to 0. The ads will be shown but without adMode (same behavior in web).

cacel:
@rcvrgs Both tickets are similar. Basically when the video is finished (with ads or without them) and after scrubbing, the button replay starts from where you finished to scrub. 
This PR, when the endScreen is shown and the user uses the scrubber bar and clicks replay button, the content will start from the beginning.

akshaydch11 (approved):

pilievOoyala:
@rcvrgs Can you re-review?  I will address the issue about replay in a separate thread.

DmitriiMaslov:
@cacel do we need to move it to DEV branch? @SergeyBicarte  @bychkovdmitry @orobles-ooyala  Please review

DmitriiMaslov:
The related ticket marked as done. PR is too old to merge it into Candidate so I am closing it.

PR #748, opened 2018-09-07
[PLAYER-4337] Convert negative seek values to 0

brol1dev:
Some devices try to seek to a negative value. Converting those values to 0.


orobles-ooyala:
If this is an iOS only issue it LGTM

orobles-ooyala (approved):

brol1dev:
@orobles-ooyala it is only for iOS

PR #750, opened 2018-09-14
Fix internal dependency vulnerabilities

brol1dev:
debug and cryptiles are internal dependencies that we don't use explicitly but are used by some of our dependencies internally. Added an explicit dependency with versions that don't have vulnerabilities reported by github.


aldigun:
LGTM.

akshaydch11:
LGTM

PR #753, opened 2018-09-19
[PLAYER-4448] Added parser for number arrays

SZadorozhnyiEpam:
Basically, to pass initial params(skin-config.json) to the skin we have to create a bundle with values.
For that, we have BundleJSONConverter.java in a native-skin layer of Android.
But since an array of numbers in skin-config appears for the first time we don't have conversion logic.
The logic was added.

I'm using String.valueOf(Object obj) because we know that every value in an array are numbers and all number subclasses have great toString(). So we can avoid a lot of if-else with checks of a concrete type.


SergeyBicarte (approved):
LGTM

brol1dev (approved):
LGTM, would be great if you add unit testing to this method.

PR #758, opened 2018-09-26
[PLAYER-4498][PLAYER-4499] Update skin-config submodule

brol1dev:
Removed playback speed rate button in skin-config so that it doesn't appear by default.


akshaydch11:
LGTM. Merging with 1 approval as it is needed to create build.

cacel:
LGTM

PR #763, opened 2018-10-17
 [PLAYER-4465] Crash when playback speed list contains empty values

bychkovdmitry:
Android json parser treats json arrays like [1, 2,, 3] as [1, 2, null, 3]. This solution simply removes redundant null values from the array.


sdk/android/skin/src/main/java/com/ooyala/android/skin/configuration/SkinConfigManager.java
```diff
@@ -0,0 +1,162 @@
...
+
+  public boolean getShowUpNext() throws JSONException {
+    return getUpNextObject().getBoolean(CONFIG_UP_NEXT_SHOW);
+  }
+
+  public boolean getShowUpNextOrDefault() {
+    try {
+      return getShowUpNext();
+    } catch (JSONException e) {
+      DebugMode.logE(TAG, "Up Next Parse Failed, default not showing Up Next");
```
brol1dev: Change the wording of the comment here. I think it is better to add something like: `"Up Next parse failed. The player will not show the up next widget."`

sdk/android/skin/src/main/java/com/ooyala/android/skin/OoyalaSkinLayoutController.java
```diff
@@ -169,33 +153,34 @@ public OoyalaSkinLayoutController(Application app, OoyalaSkinLayout layout, Ooya
...
+      ReactUtil.setJSServer(skinOptions.getReactJSServerHost(), context);
     }
+    initConfig(context, skinOptions);
+    initSkinViews(app, skinLayout, skinOptions);
+  }
+
+  private void initConfig(Context context, SkinOptions skinOptions) {
+    JSONObject configJson = AssetUtil.loadJsonAsset(context, skinOptions.getSkinConfigAssetName());
+    configManager = new SkinConfigManager(configJson);
+    configManager.removeNullsFromPlaybackArray();
```
AlinaVoronkovaEpam, line 173: I don't think it will work for iOS. Those changes should be added to React Native.

AlinaVoronkovaEpam (changes requested):

brol1dev (changes requested):
Just added one simple comment on the code comment. I like how you refactored the code, makes it simpler to read.

However, I'll say this PR is not complete yet because it is only working for Android. We should add the logic to handle iOS too or else the issue will still happen on iOS.

bychkovdmitry:
First of all, I’d like to explain what’s happening on the Android side now when we change playbackSpeed to [0.5,, 1, 1.25, 1.5, 2]:

- Android json parser parses this array to JSONArray object with the following contents: [0.5, null, 1, 1.25, 1.5, 2] 

-  Android json parser parses this array to JSONArray object with the following contents: [0.5, null, 1, 1.25, 1.5, 2] - BundleJSONConverter.convertToBundle() is used to traverse config tree. It throws and exception (which is swallowed) because it is not designed to handle null values

- We
end up passing launchOptions = null to react native, which leads to misleading exceptions like “property startScreen.promoImageSize is undefined”

bychkovdmitry:
First of all, let me explain what’s happening on the Android side now when we change playbackSpeed to [0.5,, 1, 1.25, 1.5, 2]: 
- Android json parser parses this array to JSONArray object with the following contents: [0.5, null, 1, 1.25, 1.5, 2] .
- BundleJSONConverter.convertToBundle() is used to traverse config tree and build a bundle for react native. It throws and exception (which is swallowed) because it was not designed to handle null values.  
- We end up passing launchOptions = null to react native, which leads to misleading exceptions like “property startScreen.promoImageSize is undefined”. 

Why I don’t want to fix this on the react native side:
 - Anyway we need to fix issue on Android side (and possibly on IOS side too) making some dramatic changes to BundleJSONConverter logic. 
- The config should be usable on Android side too. With null values it is flawed.
- We do not want to hide config problems (fail fast).

In my opinion config logic is terribly flawed now. Nobody knows which values are required and which are optional (a wiki page with a table should be created where it should be clearly stated which options are required and which are not, which option values are possible and which are not, what are the defaults for the absent options). Config should be checked for the presence and correctness of the values. If one one the values is absent/incorrect we should throw an exception right away (fail fast).

sdk/android/skin/src/main/java/com/ooyala/android/skin/OoyalaSkinLayoutController.java
bychkovdmitry: Alina, please have a look at my comment below.

SergeyBicarte (approved):
LGTM

SergeyBicarte (approved):
LGTM

SergeyBicarte (approved):
LGTM

SergeyBicarte (approved):
LGTM

SergeyBicarte (approved):
LGTM

SergeyBicarte (approved):
LGTM

brol1dev:
agree with moving the logic to react native @bychkovdmitry , feel free to do it. At the end of the day what we want is support for both iOS and Android.

AlinaVoronkovaEpam (approved):

PR #764, opened 2018-10-23
[Player 4564] Buttons from control panel doesn't move to more options panel if player size is too small

SergeyBicarte:
In skin.json config we have options for buttons from control panel :
"whenDoesNotFit":"moveToMoreOptions"
"whenDoesNotFit":"keep"
along with option "location":"moreOptions" (to move button to more options panel without any conditions). That is not working currently (except "location":"moreOptions" works for buttons "share" and "discovery" only). Purpose of this is resolving that issue


sdk/react/constants.js
```diff
@@ -12,7 +12,7 @@ UI_SIZES = {
 const BUTTON_NAMES = { // must match Objective-C code expectations & skin.json.
   PLAY: 'Play',
   PLAY_PAUSE: 'PlayPause',
-  FULLSCREEN: 'Fullscreen',
+  FULLSCREEN: 'fullscreen',
```
maxkupetskii: You can't do like that. It'll broke an iOS skin sdk

maxkupetskii (changes requested):

sdk/react/constants.js
SergeyBicarte: I got it, yes, but we need align these names in BUTTON_NAMES , actually they should be same as used in skin.json and i don't like change it in skin.json because it used on web also

sdk/android/skin/src/main/java/com/ooyala/android/skin/OoyalaSkinBridgeEventHandlerImpl.java
```diff
@@ -24,7 +24,7 @@
   private static final String BUTTON_SHARE = "Share";
   private static final String BUTTON_REWIND = "rewind";
   private static final String BUTTON_SOCIALSHARE = "SocialShare";
-  private static final String BUTTON_FULLSCREEN = "Fullscreen";
+  private static final String BUTTON_FULLSCREEN = "fullscreen";
```
brol1dev: why changing to lowercase?

sdk/react/constants.js
SergeyBicarte: For now (until we align all button names in some future)  i reverted it to "Fullscreen" with some changes for react native to work with that "Fullscreen" labeled button.

sdk/android/skin/src/main/java/com/ooyala/android/skin/OoyalaSkinBridgeEventHandlerImpl.java
SergeyBicarte: Look at comments above, we need align names in BUTTON_NAMES constants.js with names in "mobileContent" section of skin.json.

maxkupetskii:
Please, resolve conflicts

SergeyBicarte:
Conflicts resolved

sdk/react/controlBar.js
```diff
@@ -185,7 +185,7 @@ class ControlBar extends React.Component {
         iconTouchableStyle: styles.iconTouchable,
         durationString: this.getDurationString()
       },
-      fullscreen: {
+      Fullscreen: {
```
maxkupetskii, line 188: Why do we need this capitalized?

SergeyBicarte: In shortly, because we have FULLSCREEN: 'Fullscreen' in constants.js (BUTTON_NAMES) and that value is used in a iOS and Android bridges. 
I a long term , because we need aligned button names in all parts : bridge, react native code, skin.json. And i just started from "fullscreen" button because fix is need it to be aligned.

SergeyBicarte: In constants.js file at "const BUTTON_NAMES"  line, you can see self-explain comment :
"must match Objective-C code expectations & skin.json." , right now it is not matched. 
In skin.json it is "fullscreen" , in Android and iOS bridge parts it is "Fullscreen". This fix trying resolve it at least for "fullscreen" button.

AlinaVoronkovaEpam (approved):

maxkupetskii (approved):

PR #765, opened 2018-10-26
[PLAYER-4649] Fix for black screen on Android 8+ on skin layout

SergeyBicarte:
We need use requestLayout() methods instead of forceLayout() (to be sure views with changed size measured properly)


bychkovdmitry (approved):
LGTM

AlinaVoronkovaEpam (approved):

PR #766, opened 2018-10-29
[Player 4572] TalkBack should announce the current playback progress when seek bar is selected

SergeyBicarte:
Added (Android only) current progress % to string for scrubberBarAccessibilityLabel.


brol1dev:
How does it work on iOS currently? What's the plan to have both platforms in parity?

SergeyBicarte:
> 
> 
> How does it work on iOS currently? What's the plan to have both platforms in parity?

iOS are using platform specific  < Slider > control  as progress bar and it's already fine with announce the current playback progress. That bug is only on android react native, because of using generic  <Animated.View> control as progress bar

bychkovdmitry (approved):
LGTM

AlinaVoronkovaEpam (approved):

PR #778, opened 2018-11-15
[PBA-6587] Hide control bar when playing live videos

brol1dev:
Live assets were not hiding the control bar without interacting with it when the player is playing the asset for the first time.


ipadilla4 (approved):
LGTM

PR #791, opened 2018-12-07
[PLAYER-2203] Fixed split screen fullscreen button switch

isheftel-ooyala:
In multiWindowMode when we try to change systemUiVisibility not always this method works. Lower window in vertical orientation can't gain access to statusbar.
But anyway we need to show current state of fullscreen button icon because you can rotate the phone in split-screen mode and icons must be the same in both orientations.
So new callback was created only to change icon in react in multiWindowMode.


cacel (approved):
LGTM

SergeyBicarte (approved):
LGTM

PR #793, opened 2018-12-12
[PLAYER-4876] Fixed scroll view in audio mode

isheftel-ooyala:
In audio mode was enabled horizontal orientation of scrollview items in case width > height.
This fixes unresponsive scrollview, makes design nicer and will return vertical orientation of scrollview if we decide to make player height bigger.


MarkEPAM (approved):
LGTM

SergeyBicarte (approved):
LGTM

PR #798, opened 2018-12-18
[PLAYER-4925_4926] Fixed buttons on seekend

isheftel-ooyala:
Fixed rewind and skip buttons behavior according to this acceptance criterias:
https://jira.corp.ooyala.com/browse/PLAYER-4925
https://jira.corp.ooyala.com/browse/PLAYER-4926


MarkEPAM (changes requested):
On iOS skip controls don't work after the replay state, also `replay` button does't change icon to `playing` after the same state

isheftel-ooyala:
Why such  a difference on different platforms, on Android everything works fine only after this changes.
And before these changes problems were only on Android.
Maybe check for platform, what do you think?

maxkupetskii (approved):

SergeyBicarte (approved):
LGTM

PR #802, opened 2018-12-21
[PLAYER-3881] "Off" is not highlighted in CC list if it doesn't contain same language as was selected in previous video

AlinaVoronkovaEpam:
CC: set selected CC language as “off”, if the language selected previously is not included in the list of available CC for the current video.


SergeyBicarte (approved):
LGTM

alinag (approved):

ipadilla4 (approved):

PR #807, opened 2018-12-25
[PBA-6623] Added an ability to override jsbundle file

alinag:
Changed order in reactInstanceManager builder to avoid overwriting overriden file


sdk/android/skin/src/main/java/com/ooyala/android/skin/OoyalaSkinLayoutController.java
```diff
@@ -185,8 +185,8 @@ private void initSkinViews(Application app, OoyalaSkinLayout l, SkinOptions skin
     rootView = new ReactRootView(l.getContext());
     _reactInstanceManager = ReactInstanceManager.builder()
         .setApplication(app)
-        .setBundleAssetName(skinOptions.getBundleAssetName())
         .setJSBundleFile("assets://index.android.jsbundle")
+        .setBundleAssetName(skinOptions.getBundleAssetName())
```
SergeyBicarte, line 188: What point left here .setJSBundleFile("assets://index.android.jsbundle") call? If you are fearing  skinOptions.getBundleAssetName() returned something wrong i suggest modify getBundleAssetName() body:
```java
  public String getBundleAssetName() {
    if (bundleAssetName == null) {
      return "index.android.jsbundle";
    }
    return bundleAssetName;
  }
```

alinag: Removed that line.  `index.android.jsbundle`  is default value for `bundleAssetName`, so it should work fine without this modification

PR #809, opened 2018-12-25
[PLAYER-4968] [Audio only] Skin changes for live audio

isheftel-ooyala:
Skin changes for live audio according to this design
https://jira.corp.ooyala.com/browse/PLAYER-4427

Audio live streams are now not available, to test you can emulate on live video
1) embed code - 51Yzd1ZDE6cVwY3Brwqt0CLWfQzdpmLm
2) In sample app disable audio only flag (on Android in Options Builder in AudioOnlyPlayerActivity.class)
3) in bridge set "contentType" to "Audio" (on Android in CurrentItemChangedNotificationHandler.class)
4) for seek changes do the same in SeekStartedNotificationHandler.class and SeekCompletedNotificationHandler.class

On real live audio samples playhead position is much further to the end than in video samples, so I've made live region at 99% comparing to 95% in video.
So don't pay attention that we're not live the majority of time emulating on video.


SergeyBicarte (approved):
LGTM

maxkupetskii (approved):

PR #811, opened 2018-12-27
[PLAYER-4945] Refactored resizable player

isheftel-ooyala:
This request maybe will be deleted because of incorrect PR history, it's temporary now


zelibobla (changes requested):
Attempt to merge history that we just eliminated.

PR #827, opened 2019-01-21
[PLAYER-4998] Bottom overlay enable param from skin JSON is handled

AlinaVoronkovaEpam:
The following part of skin.json manages hiding/showing the bottom control panel:
```  
"controlBar": {
    "enabled": value,
    "volumeControl": {
      "color": ""
```
**value** could be `true`, `false`
if the param is not set, the bottom control panel shows up by default.


MarkEPAM (approved):
LGTM

SergeyBicarte (approved):
LGTM

PR #832, opened 2019-01-30
[PBA-6620] - Time controlBar asset > 1hour

ipadilla4:
We had some issues when displaying playhead for live assets, so this PR modifies how playhead looks for live asset, adding "LIVE" label instead of actual playhead.


maxkupetskii (approved):

cacel (approved):
LGTM

PR #836, opened 2019-02-06
[Player 4762] Broken fast forwarding/rewind animation while buffering

oSulyma:
Problem: Animation for activity indicator doesn't appear while video is buffering.
Problem occurs because notification for player's state `OOOoyalaPlayerStatePlaying` arrives to early, while buffering hasn't finished and playing is not started. 

The activity must be stopped in two cases: after fast-forward rewinding  and after player's initial launching.
To resolve problem in first case activity  is stopped when notification `OOOoyalaPlayerBufferingCompletedNotification` arrives. 
To handle second case when activity must be stopped after player's initial launching, was added flag that indicates if class is waiting for BufferingCompleted event 

Unit tests not included.


ipadilla4 (approved):

cacel (approved):
LGTM

PR #842, opened 2019-02-12
[PLAYER-5184] Fix 'not in ui thread' crashes

atopilskii:
Fix 'not in ui thread' crash after seek in cast mode for skin sample.


sdk/android/skin/src/main/java/com/ooyala/android/skin/OoyalaSkinBridgeEventHandlerImpl.java
```diff
@@ -120,8 +120,14 @@ public void handleRewind() {
...
-    float percent = (float) percentValue;
-    _player.seekToPercent(percent);
+    final float percent = (float) percentValue;
+    runOnUiThread(new Runnable() {
+      @Override
+      public void run() {
+        _player.seekToPercent(percent);
+      }
+    });
+
```
SergeyBicarte, line 130: Please check don't we need add same fix for handleRewind method also?

atopilskii: @SergeyBicarte no, it is not necessary for handleRewind

isheftel-ooyala (approved):

PR #844, opened 2019-02-13
[PLAYER-5195] Add handling cast list device 

atopilskii:
Add handling cast list device and passing selected device in sdk

### DIFF
--- a/sdk/android/skin/src/main/java/com/ooyala/android/skin/BridgeMessageBuilder.java
+++ b/sdk/android/skin/src/main/java/com/ooyala/android/skin/BridgeMessageBuilder.java
@@ -14,6 +14,7 @@ import com.ooyala.android.OoyalaException;
 import com.ooyala.android.OoyalaPlayer;
 import com.ooyala.android.SeekInfo;
 import com.ooyala.android.captions.ClosedCaptionsStyle;
+import com.ooyala.android.item.CastMediaRoute;
 import com.ooyala.android.item.Video;
 import com.ooyala.android.playback.PlaybackNotificationInfo;
 import com.ooyala.android.player.exoplayer.multiaudio.AudioTrack;
@@ -48,7 +49,7 @@ public class BridgeMessageBuilder {
     Double playhead = player.getPlayheadTime() / 1000.0;
 
     WritableMap params = Arguments.createMap();
-    params.putDouble("duration", duration );
+    params.putDouble("duration", duration);
     params.putDouble("playhead", playhead);
     params.putArray("availableClosedCaptionsLanguages", languages);
     params.putArray("cuePoints", cuePoints);
@@ -322,9 +323,8 @@ public class BridgeMessageBuilder {
   }
 
   /**
-   * @return Multi audio params.
-   *
    * @param audioTracks The list of available audio tracks for the current asset.
+   * @return Multi audio params.
    */
   public static WritableMap buildMultiAudioParams(List<AudioTrack> audioTracks) {
     WritableMap params = Arguments.createMap();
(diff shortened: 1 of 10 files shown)


sdk/android/skin/src/main/java/com/ooyala/android/skin/BridgeMessageBuilder.java
```diff
@@ -340,10 +340,31 @@ public static WritableMap buildMultiAudioParams(List<AudioTrack> audioTracks) {
...
+
+  /**
+   * @param castMediaRoutes The set of available devices for casting.
+   * @return list available devices.
+   */
+  public static WritableMap buildCastDeviceListParams(Object castMediaRoutes) {
+    WritableMap params = Arguments.createMap();
+    WritableArray castDeviceNames = Arguments.createArray();
+    WritableArray castDeviceIds = Arguments.createArray();
+
```
AlinaVoronkovaEpam, line 352: Init `castDeviceNames ` and `castDeviceIds` into `if (castMediaRoutes instanceof Set)` block

sdk/android/skin/src/main/java/com/ooyala/android/skin/OoyalaSkinBridgeEventHandlerImpl.java
```diff
@@ -159,6 +158,20 @@ public void onLanguageSelected(ReadableMap parameters) {
...
 
+  @Override
+  public void onCastDeviceSelected(ReadableMap parameters) {
+    final String castDeviceName = parameters.getString("castDeviceName");
+    final String castDeviceId = parameters.getString("castDeviceId");
+    if (_player != null && _player.getCurrentItem() != null) {
+      runOnUiThread(new Runnable() {
+        @Override
+        public void run() {
+          _player.selectRoute(new CastMediaRoute(castDeviceId, castDeviceName));
```
AlinaVoronkovaEpam, line 169: lets use `castDeviceName` into `run()`

sdk/android/skin/src/main/java/com/ooyala/android/skin/notification/common/CastPressedNotificationHandler.java
```diff
@@ -0,0 +1,29 @@
...
+
+  @Override
+  public String getNotificationName() {
+    return OoyalaPlayer.CAST_DEVICES_AVAILABLE_NOTIFICATION_NAME;
+  }
+
+  @Override
+  public void handle(OoyalaNotification notification) {
+    WritableMap params = BridgeMessageBuilder.buildCastDeviceListParams((Set<CastMediaRoute>) notification.getData());
+    layoutController.sendEvent(getNotificationName(), params);
```
AlinaVoronkovaEpam, line 27: Do we need to check `notification.getData()` on null?
I would add a log warning message in that case.

atopilskii: inside `buildCastDeviceListParams(...)` we do this cheking

atopilskii: so, it is not necessary

sdk/android/skin/src/main/java/com/ooyala/android/skin/OoyalaSkinBridgeEventHandlerImpl.java
atopilskii: I think, that parse map on UI is bad idea.

sdk/android/skin/src/main/java/com/ooyala/android/skin/BridgeMessageBuilder.java
atopilskii: it is bad idea, in my point of view. Because in my variant if castMediaRoutes is null or incorrect - we will pass empty lists in skin. In your variant we will pass nulls.

PR #855, opened 2019-03-06
[PLAYER-5316] Added calculated string as accessibilityLabel to _renderDefaultProgressBar()

SergeyBicarte:
_renderDefaultProgressBar method should fill accessibilityLabel with correct value on every call, this change proposing fill it with  scrubberBarAccessibilityLabel calculated in _renderCompleteProgressBar()

### DIFF
--- a/sdk/react/bottomOverlay.js
+++ b/sdk/react/bottomOverlay.js
@@ -251,14 +251,13 @@ const BottomOverlay = createReactClass({
       if ((previousAnnouncing === 0 || currentAnnouncing - previousAnnouncing > accessibilityProgressDelay)
               && currentPercent !== VALUES.MAX_PROGRESS_PERCENT) {
         previousAnnouncing = currentAnnouncing;
-        return this._renderDefaultProgressBar(playedPercent, scrubberBarAccessibilityLabel)
       } else {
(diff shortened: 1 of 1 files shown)


AlinaVoronkovaEpam (approved):

sdk/react/bottomOverlay.js
```diff
@@ -251,14 +251,13 @@ const BottomOverlay = createReactClass({
...
         previousAnnouncing = currentAnnouncing;
-        return this._renderDefaultProgressBar(playedPercent, scrubberBarAccessibilityLabel)
       } else {
         if (Platform.OS === 'android' && currentPercent === VALUES.MAX_PROGRESS_PERCENT && previousAnnouncing !== 0) {
           AndroidAccessibility.announce(scrubberBarAccessibilityLabel);
           previousAnnouncing = 0
         }
-        return this._renderDefaultProgressBar(playedPercent, "")
       }
+      return this._renderDefaultProgressBar(playedPercent, scrubberBarAccessibilityLabel)
```
atopilskii, line 260: Looks good, but I have one question: I see that early in case when 
'currentPercent === VALUES.MAX_PROGRESS_PERCENT' is true, the method _renderDefaultProgressBar is called with empty string, that prevent playing Talkback in the end of video. 
Could you clarify, did you check this case? (end of video)

atopilskii (approved):

PR #862, opened 2019-03-22
[PLAYER-5159] Fix for If stream only includes Closed Captions, the CC…

SergeyBicarte:
It shows "AudioAndCC icon in case there >1 audio tracks and "CC" icon in other case


loginov-rocks (approved):
LGTM

ooyala-amohnatkin (approved):

PR #863, opened 2019-03-25
[PLAYER-1934] Chapter Markers MVP

loginov-rocks:
This PR introduces chapter markers feature implemented at MVP level. It's dependent on iOS and Android core SDK implementation and can be merged taking into consideration it won't work until core SDKs updated.

Implementation contains Text and Icon markers, progress bar overlays as well as containers for them. Markers data passed from the skin down to the `BottomOverlay` component where containers rendered.

`BottomOverlay` component and some other components were refactored in places touched to fix lint errors. However, the logic kept the same, since it was out of the scope of this task.

In `package.json`:

- Package name changed to match npm environment naming
- Unused runtime dependencies removed
- Dev dependencies fixed to match RN v0.58
- Flow added

Common files in the root React dir like `.flowconfig`, `.gitattributes` and some others pulled from the fresh installation of v0.58.

**New components and refactored `BottomOverlay` component placed according to the new file structure that will be introduced soon. Please pay attention to the logic rather than to the lint errors or Flow errors (there are none in `src` dir, trust me).**

Tests: test suit is broken so the code can be covered after the test suit fix, which can be done as a separate task.

Known issues:

- The panel showing what will be played next is floating over the markers container if present. UX consultation needed.
- Triangle at the bottom of markers are shorter than expected on iOS. Fixed in v0.58.4 https://github.com/facebook/react-native/pull/23402

Below are the samples to show what I mean:

Screen Shot 2019-03-25 at 7 49 54 PM
Screen Shot 2019-03-25 at 8 16 07 PM


sdk/react/src/markers/MarkersContainer/MarkersContainer.js
```diff
@@ -0,0 +1,45 @@
...
+  style?: ViewStyleProp,
+};
+
+const MarkersContainer = ({
+  accentColor, duration, markers, onSeek, style,
+}: Props) => (
+  markers.length > 0
+    ? (
+      // Path pointer events through the View to its children.
+      <Animated.View pointerEvents="box-none" style={[styles.root, style]}>
```
atopilskii, line 25: What do you think about extract this large block in method?

sdk/react/src/markers/MarkersProgressBarOverlayContainer/MarkersProgressBarOverlayContainer.js
```diff
@@ -0,0 +1,42 @@
...
+  markers: Array<MarkerType>,
+  style?: ViewStyleProp,
+};
+
+const MarkersProgressBarOverlayContainer = ({
+  accentColor, duration, markers, style,
+}: Props) => (
+  markers.length > 0
+    ? (
+      <View style={[styles.root, style]}>
```
atopilskii, line 23: What do you think about extract this large block in method?

sdk/react/src/markers/MarkerProgressBarOverlay/MarkerProgressBarOverlay.js
```diff
@@ -0,0 +1,41 @@
...
+
+  const style = {};
+
+  style.left = (marker.start === 'start' ? 0 : `${marker.start / duration * 100}%`);
+
+  if (marker.end) {
+    style.right = (marker.end === 'end' ? 0 : `${100 - marker.end / duration * 100}%`);
+  }
+
+  if (marker.markerColor || accentColor) {
```
atopilskii, line 30: I'm not sure, do we need this checking?

sdk/react/ooyalaSkinBridgeListener.js
```diff
@@ -1,18 +1,31 @@
...
-} from 'react-native';
-
-import {
-  CONTENT_TYPES,
-  SCREEN_TYPES,
-  OVERLAY_TYPES,
-} from './constants';
-import Log from './log';
+export default class OoyalaSkinBridgeListener {
+  static parseMarkers(markers: ?Array<string>): Array<Marker> {
```
atopilskii: I'm not sure that it is the best place for this method. What do you think about move it somewhere?

sdk/react/src/markers/IconMarker/IconMarker.js
```diff
@@ -0,0 +1,101 @@
...
+  touchColor?: ?string,
+};
+
+type State = {
+  isExpanded: boolean,
+};
+
+export default class IconMarker extends React.Component<Props, State> {
+  static defaultProps = {
+    backgroundColor: undefined,
```
ooyala-amohnatkin, line 24: Do you really need it? They are `undefined` by default

sdk/react/src/BottomOverlay/BottomOverlay.js
```diff
@@ -0,0 +1,670 @@
+// @flow
```
atopilskii, line 1: We lost git history for this file :(

sdk/react/src/markers/IconMarker/IconMarker.js
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  handlePress() {
+    const { imageUrl, onSeek } = this.props;
+    const { isExpanded } = this.state;
+
+    // Trigger seek callback if the marker has no image or has been expanded. If there is an image we have to show it
+    // first and only after the second click on that image trigger the callback.
+    if (!imageUrl || isExpanded) {
+      onSeek();
```
ooyala-amohnatkin: I think it will be safer to call it from the callback in `setState`. Now it can be called several times

sdk/react/src/markers/MarkersContainer/MarkersContainer.js
loginov-rocks: This is a functional component which actually contains only the ternary operator returning this large block, so there is no point in extracting it into a separate component.

sdk/react/src/markers/MarkersProgressBarOverlayContainer/MarkersProgressBarOverlayContainer.js
loginov-rocks: This is a functional component which actually contains only the ternary operator returning this large block, so there is no point in extracting it into a separate component.

sdk/react/src/markers/MarkerProgressBarOverlay/MarkerProgressBarOverlay.js
loginov-rocks: From the component perspective it doesn't know anything about persistence of accentColor, because it's passed not as a config, but as a prop, so it can be used without accentColor at all. At the same time markerColor is not required property of the marker, so we need to check if anything of it is present, markerColor has more priority in that case.

sdk/react/ooyalaSkinBridgeListener.js
loginov-rocks: I believe we have to parse markers into the array of JSONs when the bridge is getting it, but I agree the code of parsing can be placed somewhere else. Do you have suggestions?

sdk/react/src/markers/IconMarker/IconMarker.js
loginov-rocks: Seems so, at least we get lint errors instead.

sdk/react/src/markers/IconMarker/IconMarker.js
loginov-rocks: Probably you're right. But it's a callback for press handling, I'm not sure that it can be called several times, could you please elaborate on that?

sdk/react/src/BottomOverlay/BottomOverlay.js
loginov-rocks: Unfortunately yes, initially I just moved and committed the file, and only after that changed it, but Git lost it anyway, don't know why :(

sdk/react/ooyalaSkinBridgeListener.js
atopilskii: I'm not sure. Maybe we can create utils folder, move util into it and also create JsonParser class that will contains this method. Also in future we can add json parsers here.

loginov-rocks: JSON parser here is just a built-in global JSON.parse method, we don't need to place it anywhere. Still, the markers array parsing logic is specific to markers, so we can create such util folder inside `src/markers/`, what do you think?

atopilskii: Yes, It sounds good for me.

loginov-rocks: https://github.com/ooyala/native-skin/pull/863/files#diff-a7855b4e985cd6026a8aedd6d6c7b80b

atopilskii (approved):

sdk/react/src/markers/IconMarker/IconMarker.js
ooyala-amohnatkin: `setState` is asynchronous, so if you tap twice state will still be old. You need to pass a function to `setState` instead of an object and call `onSeek` in the callback

loginov-rocks: Yeah, but we have to onSeek before it will be updated.

loginov-rocks: Eventually I got your point, I though you're talking about the second callback in the setState. Please review.

loginov-rocks:
@maxkupetskii @oSulyma I removed playhead property introduced by the wrong conflicts resolving in previous PRs.

ooyala-amohnatkin (approved):

PR #874, opened 2019-04-04
[PLAYER-5424] Implemented setRootRecyclerView and getTagsForScrollabl…

SergeyBicarte:
There is some kind of problem when layout view of our player contained within RecyclerView, ui scrollable elements like a progress bar, volume are not getting scroll events. That changes targets to resolve that. PR introduce 2 new methods in LayoutController interface:

setRootRecyclerView - the user should call it if there is RecyclerView used as a root container for the layout of OoyalaPlayer. Currently, it is implemented only for OoyalaSkinLayoutController in SkinSDK.

getTagsForScrollableViews - must be implemented if there is need such support. This method should return an array of string tags for views within OoyalaPlayer layout view which pretends to be scrollable. Currently it is implemented only for OoyalaSkinLayoutController in SkinSDK ("seekBar", "volumeView" returning).

### DIFF
--- a/sdk/android/skin/src/main/java/com/ooyala/android/skin/OoyalaSkinLayoutController.java
+++ b/sdk/android/skin/src/main/java/com/ooyala/android/skin/OoyalaSkinLayoutController.java
@@ -43,6 +43,8 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Observable;
 
+import androidx.recyclerview.widget.RecyclerView;
+
 import static com.ooyala.android.util.TvHelper.isTargetDeviceTV;
 
 /**
@@ -54,7 +56,7 @@ import static com.ooyala.android.util.TvHelper.isTargetDeviceTV;
  * - Observation of the OoyalaPlayer to provide up-to-date state to the UI
  * - Handlers of all React Native callbacks
  */
-public class OoyalaSkinLayoutController extends Observable implements LayoutController, OoyalaSkinLayout.FrameChangeCallback, DiscoveryManager.Callback, ReactInstanceManagerActivityPassthrough, View.OnKeyListener {
+public class OoyalaSkinLayoutController extends Observable implements LayoutController, OoyalaSkinLayout.FrameChangeCallback, DiscoveryManager.Callback, ReactInstanceManagerActivityPassthrough, View.OnKeyListener, View.OnLayoutChangeListener {
   final String TAG = this.getClass().toString();
 
   private static final double MAX_CARBOARD_DIAGONAL_INCH_VALUE = 6.5;
@@ -108,6 +110,8 @@ public class OoyalaSkinLayoutController extends Observable implements LayoutCont
   private List<Pair<String, WritableMap>> queuedEvents;
   private boolean isReactMounted;
   private boolean isTargetTV;
+  private RecyclerView rootRecyclerView = null;
+  private View.OnLayoutChangeListener recyclerLayoutChangeListener = null;
 
   /**
    * Create the OoyalaSkinLayoutController, which is the core unit of the Ooyala Skin Integration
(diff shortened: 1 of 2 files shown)


sdk/android/skin/src/main/java/com/ooyala/android/skin/OoyalaSkinLayoutController.java
```diff
@@ -108,6 +110,8 @@
   private List<Pair<String, WritableMap>> queuedEvents;
   private boolean isReactMounted;
   private boolean isTargetTV;
+  private RecyclerView rootRecyclerView = null;
+  private View.OnLayoutChangeListener recyclerLayoutChangeListener = null;
```
AlinaVoronkovaEpam: `recyclerLayoutChangeListener ` has been never used.

sdk/android/skin/src/main/java/com/ooyala/android/skin/OoyalaSkinLayoutController.java
```diff
@@ -528,6 +532,48 @@ public void switchVRMode(VrMode vrMode) {
     }
   }
 
+  @Override
+  public void onLayoutChange(View v, int left, int top, int right, int bottom, int oldLeft, int oldTop, int oldRight, int oldBottom) {
+    if (rootRecyclerView == null) {
+      return;
+    }
+    for (String tag: getTagsForScrollableViews()) {
```
AlinaVoronkovaEpam, line 542: Let's add `check on null` for `getTagsForScrollableViews()` and smth like this in that case: `DebugMode.LogW(TAG, "Warning, tags array is not initialized")`

sdk/android/skin/src/main/java/com/ooyala/android/skin/OoyalaSkinLayoutController.java
```diff
@@ -528,6 +532,48 @@ public void switchVRMode(VrMode vrMode) {
...
+            return false;
+          }
+          rootRecyclerView.requestDisallowInterceptTouchEvent(true);
+          int action = event.getActionMasked();
+          if (action == MotionEvent.ACTION_UP) {
+            rootRecyclerView.requestDisallowInterceptTouchEvent(false);
+            view.performClick();
+          }
+          return false;
+
```
AlinaVoronkovaEpam: extra line

sdk/android/skin/src/main/java/com/ooyala/android/skin/OoyalaSkinLayoutController.java
```diff
@@ -528,6 +532,48 @@ public void switchVRMode(VrMode vrMode) {
...
+        });
+      }
+    }
+  }
+
+  @Override
+  public void setRootRecyclerView(RecyclerView recyclerView) {
+
```
AlinaVoronkovaEpam: extra line here

AlinaVoronkovaEpam (changes requested):

sdk/android/skin/src/main/java/com/ooyala/android/skin/OoyalaSkinLayoutController.java
SergeyBicarte: yes, forgot to remove it after some testing , done

sdk/android/skin/src/main/java/com/ooyala/android/skin/OoyalaSkinLayoutController.java
SergeyBicarte: Check added, don't agree we need log message here. There is no point to implement setRootRecyclerView with null returned in getTagsForScrollableViews and if it return null after all , its better does not slow callback onLayoutChange() with logging call.

sdk/android/skin/src/main/java/com/ooyala/android/skin/OoyalaSkinLayoutController.java
SergeyBicarte: done

sdk/android/skin/src/main/java/com/ooyala/android/skin/OoyalaSkinLayoutController.java
SergeyBicarte: done

sdk/android/skin/src/main/java/com/ooyala/android/skin/OoyalaSkinLayoutController.java
AlinaVoronkovaEpam: Ok. agree.

AlinaVoronkovaEpam (approved):

PR #879, opened 2019-04-08
[PLAYER-5447] Refactor JSON styles to JS modules exporting StyleSheet object

loginov-rocks:
- JSON has been renamed to JS without changes in the first commit so the Git history should not be lost after merge.
- `Utils.getStyles()` removed to prevent recalculating styles and increase separation, used styles was added to the appropriate styles.
- Every styles module exports StyleSheet object by default, so we don't need to import `StyleSheet` from `react-native` anywhere except styles definitions.
- I also sorted imports in files that were touched.


maxkupetskii (approved):

sdk/react/src/shared/ItemSelectionScrollView/ItemSelectionList/ItemSelectionList.styles.js
```diff
@@ -0,0 +1,12 @@
+// @flow
+
+import { StyleSheet } from 'react-native';
+
+export default StyleSheet.create({
+  column: {
+    //
```
ooyala-amohnatkin: do we really need these empty styles?

loginov-rocks: Those are used by https://github.com/ooyala/native-skin/pull/879/files#diff-6a62de6d02c74a360ae9ff8b478c802aR26 but they were lost. I think someone who wrote that line had something in mind...

ooyala-amohnatkin: It can be just copy/paste. Anyway, if we don't use them let's remove the dead code

loginov-rocks: Removed.

ooyala-amohnatkin (approved):

PR #880, opened 2019-04-08
[PLAYER-5450] Fix dependencies to their latest patch versions

loginov-rocks:
Dependencies fixed to their latest patch versions and synced between each other. Now you won't get any warning or errors when installing npm packages.

React v16.6.3 and Flow v0.86.0 required by React Native v0.58.6, others were updated to the latest versions.

`found 11 low severity vulnerabilities` can be fixed only with migration to the latest React Native v0.59.3

### DIFF
--- a/sdk/react/package.json
+++ b/sdk/react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@ooyala/native-skin",
-  "version": "4.44.0",
+  "version": "4.45.0",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/ooyala/native-skin.git"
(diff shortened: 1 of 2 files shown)


sdk/react/package.json
```diff
@@ -1,6 +1,11 @@
 {
   "name": "@ooyala/native-skin",
   "version": "4.44.0",
```
maxkupetskii: Should it be bumped to actual one?

loginov-rocks: v4.45.0?

ooyala-amohnatkin (approved):

sdk/react/package.json
loginov-rocks: Bumped

maxkupetskii (approved):

PR #882, opened 2019-04-10
[PLAYER-5485] Sort markers by start time

loginov-rocks:


sdk/react/src/lib/markers.js
```diff
@@ -21,5 +21,28 @@ export const parseInputArray = (markers: ?Array<string>): Array<Marker> => {
         return null;
       }
     })
-    .filter(marker => Boolean(marker));
+    .filter(marker => Boolean(marker))
+    .sort(({ start: a }, { start: b }) => {
```
ooyala-amohnatkin, line 25: maybe add a test?

loginov-rocks: @ooyala-amohnatkin okay, will do, please request changes for visibility

ooyala-amohnatkin (changes requested):
please add a test

PR #886, opened 2019-04-10
[PLAYER-5424] Added special handling for volumeView from a RN controlBar

SergeyBicarte:
Due special nature of volumeView from RN control Bar, we need handle it in a place where view instanced.


AlinaVoronkovaEpam (approved):

loginov-rocks (approved):

oSulyma (approved):
LGTM because I do not understand Java

PR #902, opened 2019-04-27
[PLAYER-5549] Playback ends abruptly once the ad starts playing

ipadilla4:
This crash issue was caused because castControls object does not always exist in config. 

Solution consists in adding a little check :)


cacel (approved):
LGTM

loginov-rocks:
Hey @ipadilla4 

If there is no config.castControls then probably skin.json doesn't happen to be updated, does it?

PR #903, opened 2019-04-28
[PLAYER-5549] Playback ends abruptly once the ad starts playing

ipadilla4:
Adding validation on ad mode. CastControls object in config is not available when an ad will play and that was causing crashes.


cacel (approved):
LGTM

paparajupavan (approved):
LGTM

PR #908, opened 2019-05-03
[PLAYER-5544] App gets crashed when Clicked on the “Previous/Next Video” icon

ipadilla4:
This solution consist in hiding Next button if there is no next asset available. Next asset is obtained by discovery results, therefore we are using that to validate it.


paparajupavan (approved):
LGTM

PR #909, opened 2019-05-03
[PLAYER-5584] [PLAYER-5585] Skip Controls are missing in Audio-Only Player

cacel:
- SkipButton component has a new validation where if the visible field is not set, it returns a null view.


oSulyma (approved):
LGTM. 
Can't launch for testing because it's Android platform

paparajupavan (approved):
LGTM

cacel:
@oSulyma I tested in both platforms.
The issue was in the native-skin, the validation didn't allow to render the component.

PR #912, opened 2019-05-09
[PLAYER-5595] Video fails to seek sometimes on trying to Seek via dragging the pointer

ipadilla4:
This was an issue when you are seeking using the Skin, the progress scrubber and the playhead would not be synchronized. This fix includes adding validation for the existence of markers, since that feature caused changes in the way we calculate position. If we are in touch state and we don't have markers we can still use playhead position for progress scrubber.


cacel (approved):
LGTM

marcosmx (approved):

paparajupavan:
LGTM

PR #913, opened 2019-05-10
[PLAYER-5599] App Crashes on trying to seek with Voice Over enabled

ipadilla4:
This fix consists in passing parameters to onValueChange method, since they were not available via React props, causing the app to crash due to undefined values.


cacel (approved):
LGTM

paparajupavan:
LGTM

paparajupavan (approved):
LGTM

PR #915, opened 2019-05-11
[PLAYER-5595] Handling onTouchCancel event in position scrubber

ipadilla4:
The issue was caused because when you end the seek event outside the component area, "onTouchEnd" event will not be fired since you are no longer in component.

This fix consists in handling "onTouchCancel" event in position scrubber, to setup correctly "touch" state variable and have accurrate scrubber position.


cacel (approved):
LGTM

PR #922, opened 2019-06-05
[PLAYER-5624] Control bar is not disappeared automatically after user replays the same asset

oSulyma:
**Ticket goal:** Control bar  should be hidden automatically by timer if asset is replayed
**How PR achieves the goal:** When -showProgressBar invoked by play->replay action from OOTVGestureManager, triggeredTime is set to 'playheadTime'. But in this time player still has previous 'playheadTime' that close to player.duration. So must be changed to zero
**Affected repo-s:** iOS-native-skin only
**Unit tests:** Unit wasn't added. Warning: Unit tests target can't be built in unknown reason.
**SampleApp for testing:** TVOSSampleApp
**Notes:** JS code wasn't affected

### DIFF
--- a/sdk/tvOS/OoyalaTVSkinSDK/UI/OOOoyalaTVPlayerViewController.m
+++ b/sdk/tvOS/OoyalaTVSkinSDK/UI/OOOoyalaTVPlayerViewController.m
@@ -329,7 +329,11 @@ static OOClosedCaptionsStyle *_closedCaptionsStyle;
 }
 
 - (void)showProgressBar {
-  self.lastTriggerTime = self.player.playheadTime;
+  
+  Float64 playheadTime = self.player.playheadTime;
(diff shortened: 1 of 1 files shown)


sdk/tvOS/OoyalaTVSkinSDK/UI/OOOoyalaTVPlayerViewController.m
```diff
@@ -329,7 +326,11 @@ - (void)updateTimeWithDuration:(CGFloat)duration playhead:(CGFloat)playhead {
 }
 
 - (void)showProgressBar {
-  self.lastTriggerTime = self.player.playheadTime;
+
```
maxkupetskii: Please, remove extra line

loginov-rocks (approved):
Does it require any changes from Android side?

oSulyma:
> Does it require any changes from Android side?
@loginov-rocks 
Not. tvOS SDK-project doesn't include React-Native part. It just lays (⁨Native-skin⁩ ▸ ⁨sdk⁩ ▸ ⁨tvOS⁩) in the same repository as iOS-SDK, but doesn't use OOyalaSkinSDK and doesn't interact with JS. I agree that it's name is confusing: `OoyalaTVSkinSDK`

maxkupetskii (approved):

PR #933, opened 2019-06-26
[PLAYER-5672] Ad timer is not decremented while ad is playing

cacel:
- For Live Assets, the playhead is set by `liveHelper.frozenSeekBackTime`. When the `AdPlaybackScreen` is set, the playhead is always the initialTime and is not changed by playheadUpdate event.
I added a new boolean flag when the AdPod started (true) and AdPod completed (false), if an ad is playing, the playhead will be updated until the adPodCompleted is fired.

### DIFF
--- a/sdk/iOS/OoyalaSkinSDK/OOSkinPlayerObserver.m
+++ b/sdk/iOS/OoyalaSkinSDK/OOSkinPlayerObserver.m
@@ -35,6 +35,7 @@
 @property (nonatomic, weak) OOOoyalaPlayer *player;
 @property (nonatomic, weak) OOReactSkinModel *ooReactSkinModel;
 @property (nonatomic) struct LiveAssetHelper liveHelper;
+@property (nonatomic) BOOL isAdPlaying;
 
 @end
 
@@ -119,6 +120,7 @@ static NSString *castManagerDidDisconnectDevice = @"castDisconnected";
   if (self = [super init]) {
     _ooReactSkinModel = ooReactSkinModel;
     _player = player;
+    _isAdPlaying = NO;
     [self addSelfAsObserverToPlayer:player];
   }
   return self;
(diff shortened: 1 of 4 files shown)


sdk/iOS/OoyalaSkinSDK/OOSkinPlayerObserver.m
```diff
@@ -460,15 +462,22 @@ - (void)bridgeAdTappedNotification:(NSNotification *)notification {
 }
 
 - (void)bridgeAdPodStartedNotification:(NSNotification *)notification {
+  _isAdPlaying = YES;
```
oSulyma: Please, use self.isAdPlaying for all cases of using isAdPlaying until it's not setter implementation and not init method

sdk/iOS/OoyalaSkinSDK/OOSkinPlayerObserver.m
```diff
@@ -460,15 +462,22 @@ - (void)bridgeAdTappedNotification:(NSNotification *)notification {
...
   [self.ooReactSkinModel sendEventWithName:notification.name body:nil];
 }
 
 - (void)bridgeAdPodCompleteNotification:(NSNotification *)notification {
-  Float64 duration = self.player.duration;
-  Float64 playhead = self.player.playheadTime;
+  _isAdPlaying = NO;
```
oSulyma: The same

loginov-rocks (commented):
JS side fixed

loginov-rocks (approved):

oSulyma (approved):
iOS-part is ok

sdk/iOS/OoyalaSkinSDK/OOSkinPlayerObserver.m
oSulyma: Done

sdk/iOS/OoyalaSkinSDK/OOSkinPlayerObserver.m
oSulyma: Done

PR #936, opened 2019-07-08
[PLAYER-2645] Switch off fullscreen mode if error occurred

loginov-rocks:
If user is in the fullscreen mode and error occurs, then user doesn't have a way to exit from this mode, because Error view doesn't use control bar with fullscreen switching widget. So if error occurs, fullscreen mode will be switched off automatically.


cacel (approved):
LGTM

maxkupetskii (approved):

PR #937, opened 2019-07-08
[PLAYER-2242] Drop duration and playhead when ad starts

loginov-rocks:
New ad comes before duration and playhead get updated, so when switching from one ad to another the previous state can be rendered, including Skip button showing. So I tried to fix appropriate ticket by dropping duration and playhead state to the initial one.


cacel (approved):
LGTM

maxkupetskii (approved):

PR #943, opened 2019-07-18
[PBA-6779] Added proper method for getting activity instance

SergeyBicarte:
getContext() could return instance of ContextWrapper at some conditions, in that case old code produce crash, that fix introduce proper method to get activity for view.


maxkupetskii (approved):

loginov-rocks (approved):

AlinaVoronkovaEpam (approved):

oSulyma (approved):
LGTM

PR #945, opened 2019-07-22
[PLAYER-4346] Wrap ad playback screen with safe area to avoid iOS notch

loginov-rocks:
Safe area adds necessary paddings to avoid notches on iOS devices only. Reference: https://facebook.github.io/react-native/docs/0.59/safeareaview

Screen Shot 2019-07-22 at 8 27 32 PM


maxkupetskii (approved):

oSulyma (approved):
LGTM

